PR #11, opened 2018-08-20
Pull request to fix 'Makefile:127: recipe for target 'all' failed'

eyalron33:
Using Ubuntu 16.04, I encountered an error while building the library:
`Makefile:127: recipe for target 'all' failed`

The fix is to remove 'const' from line 289 in  src/aggregationinfo.cpp:
`std::set<const BLSPublicKey> pubKeysDedup;`

In addition, I added 'mkdir build' to README.md to complete the build instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ Cmake and a c++ compiler are required for building.
 ```bash
 git submodule init
 git submodule update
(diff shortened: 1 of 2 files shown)


mariano54:
thanks!

PR #19, opened 2018-09-18
Fouque tibouchi

mariano54:
Implements Fouque Tibouchi hashing as in this paper: https://www.di.ens.fr/~fouque/pub/latincrypt12.pdf and this PR: https://github.com/zkcrypto/pairing/pull/30. 

Sha256 is used instead of blake2b.


codablock:
Was about to ask what's behind "New constant time hashing to g2" and in the TODO and if it's in some way related to Fouque Tibouchi ;)

I've already noticed that the herumi libraries are using this for the BN curves and was wondering if this or something comparable is doable for BLS curves as well, looks like it is.

However, isn't this something that should be PR'd into the relic repository instead and then updated downstream?

mariano54:
Yeah i'm taking a look at making a PR to relic to see if they are interested in this. Have to integrate it into the config system, and maybe add in support for other curves, add tests etc.

PR #20, opened 2018-09-19
Use malloc instead of dynamic allocation on-stack in pairing functions

codablock:
Description from commit:
```
When these are called with thousands of pairings, dynamic allocation on
the stack results in stack overflows, leading to crashes.

Better to use heap allocation here.
```

If this is merged into this repo, I'll also create a PR into the relic project.

### DIFF
--- a/contrib/relic/src/pp/relic_pp_map.c
+++ b/contrib/relic/src/pp/relic_pp_map.c
@@ -49,10 +49,12 @@
  */
 static void pp_mil_k2(fp2_t r, ep_t *t, ep_t *p, ep_t *q, int m, bn_t a) {
 	fp2_t l;
-	ep_t _q[m];
+	ep_t *_q;
 	int i, j;
 
 	fp2_null(l);
+
+	_q = malloc(sizeof(ep_t) * m);
 
 	TRY {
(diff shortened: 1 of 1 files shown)


codablock:
Will run the tests later with valgrind and then report back

codablock:
Benchmark and tests run without errors. Valgrind also reported no issues.

I created an [issue](https://github.com/relic-toolkit/relic/issues/83) in relic to clarify if this would be allowed in upstream.

contrib/relic/src/pp/relic_pp_map.c
```diff
@@ -540,13 +564,17 @@ void pp_map_tatep_k12(fp12_t r, ep_t p, ep2_t q) {
...
+	ep_t *_p, *t;
+	ep2_t *_q;
 	bn_t n;
 	int i, j;
 
 	bn_null(n);
 
+	_p = malloc(sizeof(ep_t) * m);
+	_q = malloc(sizeof(ep_t) * m);
+	t = malloc(sizeof(ep2_t) * m);
```
mariano54: t should be ep_t and _q should be ep2_t

PR #25, opened 2018-09-19
Make AggregateSigsSimple public

codablock:


mariano54:
Two things:
1. AggregatePrivKeysSimple is already there, you can pass bool secure = false to AggregatePrivKeys
2. We should set the AggregationInfo for AggregateSigsSimple, that way you can call Verify(result) and it should just work.

codablock:
Removed AggregatePrivKeysSimple and moved the merging of aggregation infos from AggregateSigsInternal into AggregateSigsSimple

mariano54:
Thanks. Could you add a test that does a simple aggregation and verification?

codablock:
Good you asked for that. Turned out that https://github.com/Chia-Network/bls-signatures/pull/25/commits/8f3ec4a23f246a40c1c01374555ed6d22a2fc9bb is not a good thing for simply aggregated signatures when same messages sigs are aggregated.

TBH, I'm still in favor of decoupling the secure aggregation/verification and aggregation info from BLSSignature and have a specialized BLSSecureSignature for this. I believe it would make the code cleaner and easier to understand and avoid one aggregation mode influencing the other on code-level.

I also spent quite some time already reducing the impact of the bookkeeping in the case it is not needed. Decoupling it would make is a lot easier as well.

I propose something like this:
```c++
class BLSPrivateKey {
...
public:
  BLSSignature Sign(message) // simple signing
...
};
class BLSSignature {
public:
  ...
  bool Verify(pubKey, message) // simple verification
  BLSSignature Aggregate(otherSig) // simple aggregation
  static BLSSignature Aggregate(vector<BLSSignature> sigs) // batch aggregation
private:
  relic::g2_t sig;
};
class BLSSecureSignature {
public:
  ...
  BLSSecureSignature(const BLSSignature& r) // bootstrap from simple sig
  bool Verify() // secure verification
  BLSSecureSignature Aggregate(otherSig) // secure aggregation
  static BLSSecureSignature Aggregate(vector<BLSSecureSignature> sigs) // secure batch aggregation
private:
  BLSSignature sig;
  AggregationInfo aggregationInfo;
};
```

mariano54:
This also has some issues. If you call BlSSecureSignature.Aggregate(..other sigs..) of distinct messages, will it use secure or simple? The current Aggregate finds out the best approach and combines them. 

Also with the above approach, it's extremely easy to just use BLSSIganture aggregation and not realize that it's insecure for same message aggregation.. 

My preferred approach is keeping it as it is, but renaming AggregateSigsSimple to AggregateSigsInsecure or something more clear, or even just adding a boolean argument to Aggregate, which will just use simple aggregation.

codablock:
The pseudo code was unclear for BlSSecureSignature.Aggregate, it should only accept BlSSecureSignature as argument. The idea is that when you start doing secure aggregation, you always deal with objects of BlSSecureSignature and are thus forced to stay with secure aggregation.

We could also rename BLSSignature to BLSInsecureSignature so that it's always clear that you should read the docs (and whoever uses a class names BLSInsecureSignature without reading the docs behind it...well...should not be in crypto :D)

mariano54:
I just think it's too much overhead for people to have to understand two types of signatures, how they can be combined, the properties of this, etc. In the current model, there's only one aggergate method and one verify method, and it figures out what to do.

For more advanced uses, I agree we should probably expose simple and secure aggregation. However, we went through many iterations of the public API, and most of them turned out to be way too confusing.

mariano54:
Also, if messages are distinct, simple aggregation is faster and simpler, but an api user would not know  to use the simple aggregation methods.

codablock:
Sorry that I was unclear. In my example, BLSSecureSignature would still use simple/fast aggregation if it figures out that this is secure. It would basically be 100% what the current BLS::Aggregate/Verify would do.

mariano54:
I see. Still, it's a little confusing to have multiple types of signatures. For example, in the current spec, there's only one verify algorithm, and it's pretty straightforward what it does. This makes it so that you now have to different verify algorithms. 

If i'm a developer creating a signature object, and I have the serialized bytes, which one do I use? If I use BLSSignature, and then I try to aggregate it, I will be aggregating insecurely without meaning to do this, for example. 

If I need to aggregate a secure one with a simple one, which one do I cast to which? it's not clear. 

My preference at this point is to leave it as it is, but allow advanced users to use simple aggregation for those use-cases that need it.

codablock:
You'd always use the secure version. It would be clear for the developer at all times.

Using the insecure signature variant and the corresponding aggregation would be a conscious decision. You either use one variant or the other.

My main pain point is the inclusion of AggregationInfo in BLSSignature. It's a significant overhead for me even though I don't need it. I currently use a patched version of this library that simply does not update the aggregation info when aggregating. I also internally patched the library to have `bool BLS::VerifySimple(const BLSSignature &sig, const BLSPublicKey &pub, const uint8_t *hash)` which also ignores the aggregation info. This is a hack and I don't believe I should create a PR for this, simply because it results in BLSSignature objects with invalid/uninitialized aggregation infos (which is fine for me, but dangerous for others).

What about this: I start a basic refactoring and create a PR for this. Maybe it's easier to see then what I mean and why I believe this wouldn't be an additional burden on developers who should/want to use only secure aggregation/verification.

codablock:
See https://github.com/Chia-Network/bls-signatures/pull/32

mariano54:
Closing due to changes provided in https://github.com/Chia-Network/bls-signatures/pull/32

PR #26, opened 2018-09-19
Rename WORD and FIXED arguments to avoid conflicts in Windows build

codablock:
This fixes compilation errors when this library is used inside another project that is being cross-compiled for Windows. WORD is generally already defined in the Windows world as a 16 bit integer. FIXED is also already used.

I'm not sure if these should be really merged into this repository as it will make updating from upstream difficult later. I'll create an issue in relic to figure out if the same changes can be merged into it as well.


mariano54:
Will close, I agree that we should limit changes to relic until they merge them.

codablock:
FYI, see https://github.com/relic-toolkit/relic/pull/86

PR #29, opened 2018-09-21
Avoid pre-compressing points and remove the need for operator<

codablock:
operator< was the only reason why BLSSignature and BLSPublicKey had too keep internal serialized data up-to-data. This required calling of CompressPoint even if the data was never actually needed.

This PR is a set of refactorings which in the end remove the need for operator< and then finally removes the operator and the data field. 

It also includes a few other refactorings to reduce the amount of copying and serialization needed when sorting public keys and signatures. There is also a fix for a memory leak included.

I also removed the stl-like api (begin/end/size) as the classes were probably never intended to be used in a stl-compatible manner (and thus should not imply compatibility if there is none really). I had to change the python tests for this, but could not test these changes as tests currently fail on master as well.

The result of this PR is that copying and moving around signatures and public keys gets a lot faster. We rely on this to be fast internally as we use a lot of paralellism where ownership of objects is important.

https://github.com/Chia-Network/bls-signatures/commit/42f46606a474d5d31d4243632a1c80a809a4633b is a fix for relic, which I also create a [PR](https://github.com/relic-toolkit/relic/pull/85) for in relic.

### DIFF
--- a/src/aggregationinfo.cpp
+++ b/src/aggregationinfo.cpp
@@ -301,36 +301,44 @@ AggregationInfo AggregationInfo::SimpleMergeInfos(
 AggregationInfo AggregationInfo::SecureMergeInfos(
             std::vector<AggregationInfo> const &collidingInfosArg) {
     // Sort colliding Infos
-    std::vector<AggregationInfo> collidingInfos;
-    for (const AggregationInfo &info : collidingInfosArg) {
-        collidingInfos.push_back(info);
+    std::vector<size_t> sortedCollidingInfos(collidingInfosArg.size());
+    size_t pkCount = 0;
+    for (size_t i = 0; i < sortedCollidingInfos.size(); i++) {
+        sortedCollidingInfos[i] = i;
+        pkCount += collidingInfosArg[i].tree.size();
     }
     // Groups are sorted by message then pk then exponent
     // Each info object (and all of it's exponents) will be
     // exponentiated by one of the Ts
-    std::sort(begin(collidingInfos), end(collidingInfos));
+    std::sort(sortedCollidingInfos.begin(), sortedCollidingInfos.end(), [&collidingInfosArg](size_t a, size_t b) {
+        return collidingInfosArg[a] < collidingInfosArg[b];
+    });
 
-    std::vector<uint8_t*> sortedKeys;
-    for (auto &info : collidingInfos) {
-        for (auto &mapEntry : info.tree) {
-            sortedKeys.push_back(mapEntry.first);
+    std::vector<uint8_t*> msgAndPks;
+    std::vector<std::vector<uint8_t>> serPks;
+    std::vector<size_t> sortedKeys;
+    msgAndPks.reserve(pkCount);
+    serPks.reserve(pkCount);
+    sortedKeys.reserve(pkCount);
+    for (size_t i = 0; i < sortedCollidingInfos.size(); i++) {
+        for (auto &mapEntry : collidingInfosArg[sortedCollidingInfos[i]].tree) {
+            msgAndPks.emplace_back(mapEntry.first);
+            serPks.emplace_back(mapEntry.first + BLS::MESSAGE_HASH_LEN, mapEntry.first + BLS::MESSAGE_HASH_LEN + BLSPublicKey::PUBLIC_KEY_SIZE);
+            sortedKeys.emplace_back(sortedKeys.size());
         }
     }
     // Pks are sorted by message then pk
-    std::sort(begin(sortedKeys), end(sortedKeys), BLSUtil::BytesCompare80());
-    std::vector<BLSPublicKey> sortedPks;
-    for (const uint8_t* key : sortedKeys) {
-        sortedPks.push_back(BLSPublicKey::FromBytes(
-                key + BLS::MESSAGE_HASH_LEN));
-    }
+    std::sort(sortedKeys.begin(), sortedKeys.end(), [&msgAndPks](size_t a, size_t b) {
+        return memcmp(msgAndPks[a], msgAndPks[b], BLS::MESSAGE_HASH_LEN + BLSPublicKey::PUBLIC_KEY_SIZE) < 0;
+    });
 
     // Calculate Ts
     // Each T is multiplied with an exponent in one of the collidingInfos
-    relic::bn_t* computedTs = new relic::bn_t[collidingInfos.size()];
-    for (size_t i = 0; i < collidingInfos.size(); i++) {
+    relic::bn_t* computedTs = new relic::bn_t[sortedCollidingInfos.size()];
+    for (size_t i = 0; i < sortedCollidingInfos.size(); i++) {
         bn_new(computedTs[i]);
     }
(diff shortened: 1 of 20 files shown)


src/aggregationinfo.cpp
```diff
@@ -301,45 +301,53 @@ AggregationInfo AggregationInfo::SimpleMergeInfos(
...
+    sortedKeys.reserve(pkCount);
+    for (size_t i = 0; i < sortedCollidingInfos.size(); i++) {
+        for (auto &mapEntry : collidingInfosArg[sortedCollidingInfos[i]].tree) {
+            msgAndPks.emplace_back(mapEntry.first);
+            serPks.emplace_back(mapEntry.first + BLS::MESSAGE_HASH_LEN, mapEntry.first + BLS::MESSAGE_HASH_LEN + BLSPublicKey::PUBLIC_KEY_SIZE);
+            sortedKeys.emplace_back(sortedKeys.size());
         }
     }
     // Pks are sorted by message then pk
-    std::sort(begin(sortedKeys), end(sortedKeys), BLSUtil::BytesCompare80());
```
mariano54: Are we using BytesCompare80 anymore? If not we can remove it from BLSUtil

src/blspublickey.cpp
```diff
@@ -39,39 +38,25 @@ BLSPublicKey BLSPublicKey::FromG1(const relic::g1_t* pubKey) {
...
 
-size_t BLSPublicKey::size() const {
-    return PUBLIC_KEY_SIZE;
+void BLSPublicKey::Serialize(uint8_t *buffer) const {
+    BLS::AssertInitialized();
+    CompressPoint(buffer, &q);
 }
 
-const uint8_t* BLSPublicKey::begin() const {
+std::vector<uint8_t> BLSPublicKey::Serialize() const {
```
mariano54: Why do we need a second serialize method?

src/bls.cpp
```diff
@@ -572,22 +591,23 @@ bool BLS::VerifyNative(
...
 
-    for (size_t i = 0; i < pubKeys.size(); i++) {
-        pubKeys[i].Serialize(pkBuffer +
-                             i * BLSPublicKey::PUBLIC_KEY_SIZE);
-    }
-
     bn_new(order);
     g2_get_ord(order);
 
+    std::vector<uint8_t> pkBuffer;
```
mariano54: Why use a vector here instead of the array?

mariano54: array seems cleaner and more consistent with the rest of the code

src/bls.cpp
```diff
@@ -572,22 +591,23 @@ bool BLS::VerifyNative(
 }
 
 void BLS::HashPubKeys(relic::bn_t* output, size_t numOutputs,
-                      std::vector<BLSPublicKey> const &pubKeys) {
-    uint8_t *pkBuffer = new uint8_t[BLSPublicKey::PUBLIC_KEY_SIZE
-                                    * (pubKeys.size())];
+                      std::vector<std::vector<uint8_t>> const &serPubKeys,
+                      std::vector<size_t> const& sorted) {
```
mariano54: Maybe rename sorted to sortedIndeces

src/blssignature.hpp
```diff
@@ -66,30 +69,24 @@ class BLSSignature {
...
 
-    const uint8_t& operator[](size_t pos) const;
-    const uint8_t* begin() const;
-    const uint8_t* end() const;
-    size_t size() const;
-
     // Serializes ONLY the 96 byte public key. It does not serialize
     // the aggregation info.
     void Serialize(uint8_t* buffer) const;
+    std::vector<uint8_t> Serialize() const;
```
mariano54: If we want to keep both serialize, we should also do it for other classes like ExtendedPubKey, ChainCode, etc

mariano54:
Python bindings tests are passing on my machine

src/bls.cpp
mariano54: And other places where arrays are replaced by vectors

codablock: With arrays you mean dynamically allocated arrays as used before or std::array?

I prefer the use of std::vector<uint8_t> over new[] whenever possible. I actually prefer to not use new/new[] at all, simply because it is one of the main sources for memory related bugs. The memory leak that I fixed in fdf15f98ea34a8a77cc88e7b9455d4d999fa5155 was such an example. With std::vector you've generally on the secure side of things and also get a lot of flexibility (e.g. everything from std can be used, or the < which makes sorting simpler). You also have better control about ownership of buffers/arrays, as you can always pass around ownership with std::move/std::swap.

About consistency: I noticed this as well and actually started a refactoring of all places where new[] was used...but stopped in the middle as it got too much. Would you consider using std::vector in the future (for all new things) and gradually move new[] to std::vector step by step?

src/aggregationinfo.cpp
codablock: still used in SortIntoVectors

src/bls.cpp
codablock: done

src/blspublickey.cpp
codablock: 1. for convenience (otherwise you'd have to do `std::vector<uint8_t> b; b.resize(size); pk.Serialize(b.data())`
2. as a replacement when begin/end/size was needed (which I removed).

src/bls.cpp
codablock: With modern C++ (starting with C++11), the overhead is neglectable. Only when iterating byte-by-byte, you'd be able to measure a performance impact. But usually, you'd use stuff from std:: to perform operations, which all have specialized implementations to work fast on primitive data types.

It's however very important to pass vectors by reference, as otherwise you end up copying them over and over. Returning vectors by-value depends on the place it is used. If the vector is from the stack, the C++ compiler will will use return value optimization (avoiding a copy). If you return a member variable, it will be copied.

If you put vectors in maps or other vectors, always use m.emplace(k, vec) and v2.emplace_back(vec). If the vector is not required after adding it to another container, use m.emplace(std::move(vec)) and v2.emplace_back(std::move(vec)), which will basically swap (very fast) the contents of the inserted vector into the new position in the target container. Generally, always prefer emplace over push/insert (it avoids unnecessary copies).

If using vectors in vectors, make sure to v.reserve(expectedSize) in advance, otherwise it might end up copying stuff unnecessarily.

jonspock: @codablock do you have some backup for the negligible overhead of std::vector with C++11? 
Another alternative is to use std::unique_ptr. This avoids the memory bugs and ensures that the size is set and fixed for all time. Thus making the code easier to understand than using std::vector (across the board) where the size could be set at any place or time. With std::vector it can be harder to know whether additional elements have been added or not later.

mariano54:
Other c++ crypto libraries I've seen, use pointers, for example https://github.com/weidai11/cryptopp.

I'm open to switching to vectors, but I think this requires more thought and should be done in a different PR, which keeps everything consistent. @codablock could you use arrays for this PR instead? This bug fix and refactoring is still definitely useful.

codablock:
Rebased on master on added commits to revert back to uint8_t*

PR #32, opened 2018-09-26
Decouple aggregation info from signature and other refactorings

codablock:
This refactoring got quite huge. I'd recommend to review on commit level (instead of a full diff on all changes). The PR basically does the following:

1. It splits BLSSignature into BLSInsecureSignature and BLSSignature. BLSInsecureSignature is meant to be used internally only, and by developers who want to use simple/basic/insecure aggregation and verification (as I need it). BLSSignature internally just uses BLSInsecureSignature, but makes sure aggregation and verification is actually done in a secure manner. There is no actual change in the aggregation/verification logic.
2. Move nearly all methods (mostly aggregation related) from `class BLS` to the BLSPublicKey, BLSPrivateKey, BLSInsecureSignature and BLSSignature. This way the API is more object oriented and functionality is placed where it belongs instead of having everything in one class.
3. BLSPublicKey and BLSInsecureSignature now have a `Mul(bn)` method. This allowed to remove all external accesses to the group elements of these classes. This simplified a few places and made them easier to read and maintain.

Please look into the changes in README.md to see how this affects the public API. As you can see, it's basically the same and BLSInsecureSignature had no influence on the API. The only change in the API that you see are related to moving methods from `BLS` into the other classes.

The changes in test.cpp should also show that BLSInsecureSignature has no influence on how secure aggregation/verification is done with the public API.


src/blssignature.cpp
```diff
@@ -36,35 +36,134 @@ BLSSignature BLSSignature::FromBytes(const uint8_t *data) {
...
+BLSInsecureSignature::BLSInsecureSignature(const BLSInsecureSignature &signature) {
+    BLS::AssertInitialized();
+    g2_copy(sig, *(relic::g2_t*)&signature.sig);
+}
+
+void BLSInsecureSignature::GetPoint(relic::ep2_st* output) const {
+    *output = *sig;
+}
+
+BLSInsecureSignature BLSInsecureSignature::Aggregate(const BLSInsecureSignature& r) const {
```
mariano54: We can call the Aggregate of vector from this one

src/blssignature.cpp
```diff
@@ -36,35 +36,134 @@ BLSSignature BLSSignature::FromBytes(const uint8_t *data) {
...
+        throw std::string("sigs must not be empty");
+    }
+    BLSInsecureSignature result = sigs[0];
+    for (size_t i = 1; i < sigs.size(); i++) {
+        g2_add(result.sig, result.sig, *(relic::g2_t*)&sigs[i].sig);
+    }
+    return result;
+}
+
+BLSInsecureSignature BLSInsecureSignature::DivideBy(const BLSInsecureSignature& r) const {
```
mariano54: We should have divide by vector of signatures as well

src/blssignature.cpp
```diff
@@ -80,6 +80,12 @@ BLSInsecureSignature BLSInsecureSignature::DivideBy(const BLSInsecureSignature&
     return result;
 }
 
+BLSInsecureSignature BLSInsecureSignature::Mul(const relic::bn_t n) const {
```
mariano54: We are using multiplicative notation, so multiplication should refer to multiplying two group elements, and exponentiation should refer to raising a group element to a scalar. So we should rename this to something like "Raise". 

I know this is not how relic does it, they use additive notation, but it's how it is described in the academic papers.

mariano54: I also didn't implement it this way initially because of the overhead of creating new objects. I suppose, since there is no point compression going on here, that this is fast? Have you run the benchmarks?

src/blssignature.cpp
```diff
@@ -36,35 +36,134 @@ BLSSignature BLSSignature::FromBytes(const uint8_t *data) {
...
+    *output = *sig;
+}
+
+BLSInsecureSignature BLSInsecureSignature::Aggregate(const BLSInsecureSignature& r) const {
+    BLSInsecureSignature result(*this);
+    g2_add(result.sig, *(relic::g2_t*)&result.sig, *(relic::g2_t*)&r.sig);
+    return result;
+}
+
+BLSInsecureSignature BLSInsecureSignature::Aggregate(const std::vector<BLSInsecureSignature>& sigs) {
```
mariano54: Can we rename these two methods to Mul? All it's doing is multiplying the signatures together.

mariano54: It is also the opposite of divide, so it make sense

src/blssignature.cpp
```diff
@@ -60,6 +60,73 @@ void BLSInsecureSignature::GetPoint(relic::ep2_st* output) const {
     *output = *sig;
 }
 
+bool BLSInsecureSignature::Verify(const uint8_t* msg, size_t len, const BLSPublicKey& pubKey) const {
```
mariano54: I'm not super convinced on BLSInsecureSignature having such a different interface compared to BLSSignature

src/blspublickey.cpp
```diff
@@ -41,11 +42,87 @@ BLSPublicKey BLSPublicKey::FromG1(const relic::g1_t* pubKey) {
...
+    }
+    for (auto p : serPubKeys) {
+        delete[] p;
+    }
+
+    BLS::CheckRelicErrors();
+    return aggKey;
+}
+
+BLSPublicKey BLSPublicKey::Mul(const relic::bn_st* n) const {
```
mariano54: Should be exponentiation

src/blspublickey.cpp
```diff
@@ -41,11 +42,87 @@ BLSPublicKey BLSPublicKey::FromG1(const relic::g1_t* pubKey) {
...
+    BLS::AssertInitialized();
+    g1_set_infty(q);
+}
+
 BLSPublicKey::BLSPublicKey(const BLSPublicKey &pubKey) {
     BLS::AssertInitialized();
     g1_copy(q, pubKey.q);
 }
 
+BLSPublicKey BLSPublicKey::AggregateInsecure(const BLSPublicKey& r) const {
```
mariano54: Let's just remove this one, it's not too difficult to create a vector

src/blspublickey.cpp
```diff
@@ -41,11 +42,87 @@ BLSPublicKey BLSPublicKey::FromG1(const relic::g1_t* pubKey) {
...
+    BLSPublicKey ret;
+    g1_add(ret.q, q, r.q);
+    return ret;
+}
+
+BLSPublicKey BLSPublicKey::AggregatePubKeysInsecure(std::vector<BLSPublicKey> const& pubKeys) {
```
mariano54: This should be called Mul

src/blsprivatekey.cpp
```diff
@@ -92,6 +94,99 @@ BLSPublicKey BLSPrivateKey::GetPublicKey() const {
     return ret;
 }
 
+BLSPrivateKey BLSPrivateKey::AggregateInsecure(const BLSPrivateKey& r) const {
```
mariano54: Remove this one

src/blsprivatekey.cpp
```diff
@@ -92,6 +94,99 @@ BLSPublicKey BLSPrivateKey::GetPublicKey() const {
...
+    bn_new(order);
+    g1_get_ord(order);
+
+    BLSPrivateKey ret(*this);
+    relic::bn_add(*ret.keydata, *ret.keydata, *r.keydata);
+    relic::bn_mod_basic(*ret.keydata, *ret.keydata, order);
+    return ret;
+}
+
+BLSPrivateKey BLSPrivateKey::AggregatePrivKeysInsecure(std::vector<BLSPrivateKey> const& privateKeys) {
```
mariano54: Should be called Add

src/blsprivatekey.cpp
```diff
@@ -41,7 +41,7 @@ BLSPrivateKey BLSPrivateKey::FromSeed(const uint8_t* seed, size_t seedLen) {
     bn_read_bin(*skBn, hash, BLSPrivateKey::PRIVATE_KEY_SIZE);
     bn_mod_basic(*skBn, *skBn, order);
 
-    BLSPrivateKey k = BLSPrivateKey();
+    BLSPrivateKey k;
```
mariano54: What's the difference here?

src/blsprivatekey.hpp
```diff
@@ -42,6 +42,7 @@ class BLSPrivateKey {
     // Construct a private key from another private key. Allocates memory in
     // secure heap, and copies keydata.
     BLSPrivateKey(const BLSPrivateKey& k);
+    BLSPrivateKey(BLSPrivateKey&& k);
```
mariano54: What is the purpose of this?

src/blssignature.cpp
```diff
@@ -68,13 +64,12 @@ bool BLSInsecureSignature::Verify(const uint8_t* msg, size_t len, const BLSPubli
...
 bool BLSInsecureSignature::VerifyHash(const uint8_t* hash, const BLSPublicKey& pubKey) const {
-    relic::g1_t pk, gen;
+    relic::g1_t gen;
     relic::g2_t mappedHash;
     relic::gt_t e1, e2;
     g1_get_gen(gen);
-    pubKey.GetPoint(pk);
     g2_map(mappedHash, hash, BLS::MESSAGE_HASH_LEN, 0);
-    pc_map(e1, pk, mappedHash);
+    pc_map(e1, *(relic::g1_t*)&pubKey.q, mappedHash);
```
mariano54: Why are we using . notation instead of a getter?

src/extendedprivatekey.cpp
```diff
@@ -117,31 +117,16 @@ ExtendedPrivateKey ExtendedPrivateKey::PrivateChild(uint32_t i) const {
...
-
-    bn_add(*newSk, *newSk, *sk.GetValue());
-    bn_mod_basic(*newSk, *newSk, order);
-
-    uint8_t* newSkBytes = BLSUtil::SecAlloc<uint8_t>(
-            BLSPrivateKey::PRIVATE_KEY_SIZE);
-    bn_write_bin(newSkBytes, BLSPrivateKey::PRIVATE_KEY_SIZE, *newSk);
-    BLSPrivateKey skp = BLSPrivateKey::FromBytes(newSkBytes);
+    BLSPrivateKey newSk = BLSPrivateKey::FromBytes(ILeft, true);
+    newSk = newSk.AggregateInsecure(sk);
```
mariano54: Nice simplification..

src/blsprivatekey.cpp
```diff
@@ -220,21 +220,35 @@ std::vector<uint8_t> BLSPrivateKey::Serialize() const {
     return data;
 }
 
-BLSSignature BLSPrivateKey::Sign(const uint8_t *msg, size_t len) const {
+BLSInsecureSignature BLSPrivateKey::SignInsecure(const uint8_t *msg, size_t len) const {
```
mariano54: I don't like this. I prefer having a method in BLSSignature to convert it to an insecure signature. So you can just do BLSPrivateKey::Sign(...).ToInsecure()

src/blsprivatekey.hpp
```diff
@@ -48,8 +48,11 @@ class BLSPrivateKey {
 
     BLSPublicKey GetPublicKey() const;
 
+    // Insecurely aggregate multiple private keys into one
```
mariano54: These comments should be updated based on my other feedback on renaming methods

src/blssignature.cpp
mariano54: Actually, let's just remove the one signature Aggregate for now.

src/blssignature.cpp
mariano54: Actually, let's only have the vector versions of everything for now, to keep API simpler

src/blssignature.cpp
codablock: It should be fast as the compiler does return value optimizations here, meaning that it avoids unnecessary copies as much as possible. I've run the benchmarks and did not notice any negative impacts. It's actually slightly faster, but I only did a few bench runs so that might be variance.

src/blsprivatekey.cpp
codablock: Technically, you did a default-construction and assignment operations, so you had one unnecessary copy here. Practically, the compiler would have likely removed the unnecessary copy on it's own.

src/blsprivatekey.hpp
codablock: It's a move constructor, which allows the compiler to do proper return value optimization. It also speeds up all places were a move/swap could be used instead of a copy, e.g. what vectors internally do when enlarging the internal buffers.

The compiler usually creates a default move constructor, but I decided to make this explicit here as I want to be 100% sure that no unnecessary copies are done when moving/swapping.

When this constructor is called, it will swap the internal keydata with the given rvalue reference's keydata. It basically moves around ownership of keydata. You can also invoke this constructor if needed with `BLSPrivateKey newKey(std::move(oldKey))`, after which `newKey` will own the keydata and `oldKey` will be uninitialized (as it now owns the nullptr keydata). This is basically what the stl libraries do a lot and what the compiler also does to optimize stuff.

src/blssignature.cpp
codablock: I wanted to get rid of the getters completely so that the internals are fully hidden in the 3 classes (keys and sig). The Aggregate/Mul methods helped a lot with getting rid of them, as the external operations on group elements were the main reason the getters were needed.

src/blsprivatekey.cpp
codablock: That would bring us back to unnecessary bookkeeping of aggregation info when the caller doesn't need it, which was a big problem for me tbh. If done how you suggested it, it would temporarily create a BLSSignature with proper aggregation info and then throw it away.

The idea behind BLSInsecureSignature is that it is the actual internal implementation of plain signatures. As you can see in BLSSignature, verification is handed over to BLSInsecureSignature after the proper exponentiation has been performed.

I understand that you don't like having 2 different APIs here. That's why I chose to explicitly mark all low-level/insecure methods as insecure. These aren't even documented in the README.md, so a not-knowing developer wouldn't even encounter the insecure API when following the examples. If a developer encounters these insecure methods and uses them, he should be aware of what he's doing (we're warning him in the name of every method he calls...). If he still uses them, he either knows what he's doing or he probably shouldn't work on crypto projects (I know, harsh, but that's my humble opinion) because someone ignoring such obvious warning would tend to do other huge mistakes as well.

codablock:
Regarding renaming of `Aggregate` and `Mul`. As you know from our email exchanges, I was confused about the different notations (additive/multiplicative) as well :D I assume many other developers will be confused as well. That's why I believe that `Aggregate` is a better fitting method name for a public facing API, as it leaves out the confusion about different notations. At the same time external developers should usually not try to do advances math on the BLS primitives, so best is they don't know that `Aggregate` is actually multiplication or addition depending on notation (we can see it as an implementation detail). 

`Mul` is a different thing...That one should probably not be part of the public API at all. `Raise` or `Exp` fits well here as an internal API name, especially as it matches with what all the documentation/comments say ("exponentiating public keys" and so on).

There is another level of abstraction that I have in mind, which will become important if support for (non-)swapped G1 and G2 or other curves is needed. For us it's currently unclear if we actually need that as we might also decide to use your implementation as-is (BLS12-381 reversed).

But if that abstraction is needed at some point, we would have to introduce an abstraction around G1 and G2 with basic mathematical operations. This abstraction would be low-level and that's a point were proper naming in regard to notation will be important.

Regarding removal of `Aggregate` with a single parameter (non-vector versions), I will remove them now and only leave the vector versions.

codablock:
Regarding `DivideBy`, what about `Deaggregate`? This would also hide the notation and implementation details from the public API?

codablock:
Pushed additional commits, mainly containing the removal of the one-argument Aggregate versions. I only renamed `Mul` to `Exp` until you comment on my answers. I hope `Exp` instead of `Raise` is also ok for you? If not, I'll change it to `Raise` as part of my next review fixes/changes.

src/blsprivatekey.hpp
```diff
@@ -55,6 +55,9 @@ class BLSPrivateKey {
     static BLSPrivateKey Aggregate(std::vector<BLSPrivateKey> const& privateKeys,
                                    std::vector<BLSPublicKey> const& pubKeys);
 
+    // Exponentiate private key with n
```
mariano54: This is not exponentiation, it's just mutliplication

src/blsprivatekey.hpp
```diff
@@ -78,11 +75,15 @@ class BLSPrivateKey {
     // Don't allow public construction, force static methods
     BLSPrivateKey() {}
 
-    // The actual byte data
-    relic::bn_t *keydata{nullptr};
+    // Exponentiate private key with n
```
mariano54: Again, comment should say multiply not exponentiate

src/blsprivatekey.cpp
mariano54: OK, you convinced me

src/blssignature.cpp
mariano54: These pairings are not being done in the most efficient way. pc_map_sim should be used instead, look at VerifyNative. In fact, I think we should remove these methods (VerifyInsecure and VerifyInsecureHas) as VerifyInsecureAggregated does both of them. Also, can we rename it to just Verify? So there is a BLSInsecureSignature::Verify  and BLSSIgnature::Verify.

codablock:
While removing the Verify/VerifyHash methods I noticed that they are used very often in tests. Removing them would require to put the message hashing into the test cases which would bloat them up quite a bit. I left them in for now, unless you say I should remove them nevertheless and accept the bloat in test cases.

Verify still calls VerifyHash, which now calls VerifyAggregated instead of doing 2 pc_map calls.

mariano54:
Yes i prefer to remove them. I value simplicity of API at this point. Convenience methods can be added later if they are really needed.

codablock:
Added a new commit which removes the single hash/message versions of Verify and renames VerifyAggregated to Verify. I hope that all review comments are handled now and that the PR is ready for merge :)

PR #35, opened 2018-10-02
Use bls namespace , rename files and classes

jonspock:
This is done as separate commits to make diffing the files easier.
A couple of minor fixes in Cmake files in addition
Can squash to a single commit if desired


jonspock:
OK, updated README.

mariano54:
Thanks!

PR #36, opened 2018-10-03
Fix multiple errors in Chia related changes to relic if ALLOC!=AUTO

codablock:
These are mostly compilation errors and a few fixes for memory leaks.

I encountered these while trying to build and test the library with ALLOC=DYNAMIC. This PR only fixes the issues which I found in the relic source. There are much more issues with the BLS library, but I'm not going to create a PR for those...these required much more changes. And instead of fixing the issues on that level, I'd suggest to implement small C++ wrappers around relic::bn_t, relic::ep_t and relic::ep2_t which do the appropriate relic::xxx_new/relic::xxx_free calls in constructors/destructors. This way you'd catch all possible memory leaks and clean up the code as well.

I decided to create this PR so that whenever you decide to backport the `ep_map` related changes to upstream, these changes are included.

And FYI (not really related to this PR), the tests I did with ALLOC=DYNAMIC were a dead end. As expected, it got slower in single threaded environment, but my hope was that it would improve multithreaded environments, where currently parallel verification does not scale after about 4 threads on my system (with 8 cores). I assume that the issues I have with multithreading are in some way related to L1/L2 cache issues and as I had no idea how to fix this, I gave ALLOC=DYNAMIC a try to compare it ;)

### DIFF
--- a/contrib/relic/src/ep/relic_ep_map.c
+++ b/contrib/relic/src/ep/relic_ep_map.c
@@ -159,6 +159,14 @@ void ep_sw_encode(ep_t p, fp_t t) {
 /*============================================================================*/
 
 void ep_map(ep_t p, const uint8_t *msg, int len) {
+	bn_t t0;
+	bn_t t1;
+	fp_t t0p;
+	fp_t t1p;
+	ep_t p0;
+	ep_t p1;
+	bn_t k;
+
 	TRY {
 		uint8_t input[MD_LEN + 5];
 		md_map(input, msg, len);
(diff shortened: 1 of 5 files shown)


mariano54:
ALLOC dynamic and other types of allocation are going to be removed eventually in relic, so I didn't try to get it to work with those (only auto). Using other allocs requires calling ep_new, ep_null, ep_free, etc, so it also clutters up the code a lot.

In ALLOC=AUTO, most of these operations actually do nothing at all, so I just omitted them to keep the code cleaner. bn_new is an exception, in that it actually has to get called. Anyway, I think these changes make sense to keep the relic code consistent.

Issue #38, opened 2018-10-17
Aggregation Info's `=` operator calls Clear, before anything else, and segfaults.


kayabaNerve:
Line 399 of AggregationInfo tries to delete a null element and causes a segfault.

My code likely won't be too useful as it's not in C++, but here's a link to my test that is demonstrating the issue: https://github.com/EmberCrypto/BLS/blob/adabf81f18343be9007abba61bf70868b47a247f/tests/SignatureTest.nim
And my library file that's wrapping the C++:
https://github.com/EmberCrypto/BLS/blob/adabf81f18343be9007abba61bf70868b47a247f/BLS/PrivateKey.nim#L94

@jonspock looked it over with me and we both believe it to not be a problem with Nim.

jonspock:
I believe the following loops should only execute if the variables are not nulltpr/empty.
Using the operator= , there should only be a Clear() for pre-existing objects or re-code to avoid these loops (hope that is clear, if not, please ping me)

    for (auto &mapEntry : tree) {
        delete[] mapEntry.second;
        mapKeys.push_back(mapEntry.first);
    }
    for (const uint8_t* mapKey : mapKeys) {
        delete[] mapKey;
    }

kayabaNerve:
Aren't these loops reducible to the below, anyways?
```
for (auto &mapEntry : tree) {
    delete[] mapEntry.first;
    delete[] mapEntry.second;
}
```

mariano54:
Confirmed that this issue is fixed by #39?

kayabaNerve:
Yep. I confirmed it before I made the PR.

PR #39, opened 2018-10-18
Only delete tree entries when the tree isn't empty.

kayabaNerve:
Fixes https://github.com/Chia-Network/bls-signatures/issues/38

### DIFF
--- a/src/aggregationinfo.cpp
+++ b/src/aggregationinfo.cpp
@@ -394,15 +394,13 @@ AggregationInfo AggregationInfo::SecureMergeInfos(
 void AggregationInfo::Clear() {
     sortedMessageHashes.clear();
     sortedPubKeys.clear();
-    std::vector<uint8_t*> mapKeys;
-    for (auto &mapEntry : tree) {
-        delete[] mapEntry.second;
-        mapKeys.push_back(mapEntry.first);
+    if (!(tree.empty())) {
+        for (auto &mapEntry : tree) {
+            delete[] mapEntry.first;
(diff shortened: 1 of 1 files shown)


src/aggregationinfo.cpp
```diff
@@ -394,15 +394,13 @@ AggregationInfo AggregationInfo::SecureMergeInfos(
...
-    for (auto &mapEntry : tree) {
-        delete[] mapEntry.second;
-        mapKeys.push_back(mapEntry.first);
-    }
-    for (const uint8_t* mapKey : mapKeys) {
-        delete[] mapKey;
+    if (!(tree.empty())) {
+        for (auto &mapEntry : tree) {
+            delete[] mapEntry.first;
+            delete[] mapEntry.second;
```
mariano54: Are you sure it's safe to delete these bytes while iterating through the map? I can't remember why I did it the original way

kayabaNerve: I wouldn't call myself an expert on whether or not it's safe, but I have no idea why it wouldn't be. 

You had a loop that deleted the second, and pushed the first on to a vector, defined inside that scope and not used elsewhere. Then you had a loop that deleted everything in the vector. This combined the two loops and removed the vector.

If you needed to delete the second before the first due to a dependence, you could just reorder those two lines.

The loop I submitted still passes the tests and @jonspock agreed they should be equivalent.

mariano54: OK, we can merge it, since automated tests run valgrind and found no issues.

PR #43, opened 2018-10-31
Implement DHKeyExchange

codablock:
This implements primitive support for Diffie-Hellman key exchanges.

### DIFF
--- a/src/bls.cpp
+++ b/src/bls.cpp
@@ -105,7 +105,9 @@ PublicKey BLS::DHKeyExchange(const PrivateKey& privKey, const PublicKey& pubKey)
     if (!privKey.keydata) {
         throw std::string("keydata not initialized");
     }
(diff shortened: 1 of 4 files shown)


src/bls.cpp
```diff
@@ -100,6 +100,14 @@ void BLS::HashPubKeys(bn_t* output, size_t numOutputs,
     CheckRelicErrors();
 }
 
+PublicKey BLS::DHKeyExchange(const PrivateKey& privKey, const PublicKey& pubKey) {
+    AssertInitialized();
+    if (!privKey.keydata) {
+        throw std::string("keydata not initialized");
+    }
+    return pubKey.Exp(*privKey.keydata);
```
mariano54: CheckRelicErrors after the exponentiation? I think we should do this in every function just in case

PR #48, opened 2018-11-06
Threshold

AWice:


python-impl/keys.py
```diff
@@ -83,6 +87,31 @@ def from_bytes(buffer):
...
+          - commitments to your polynomial P
+          - secret_fragments[j] = P(j), to be sent to player j
+            (All N secret_fragments[j] can be combined to make a secret share.)
+        """
+        assert 1 <= T <= N
+        g1 = generator_Fq()
+        poly = [Fq(default_ec.n, RNG.randint(1, default_ec.n - 1))
+                for _ in range(T)]
+        commitments = [g1 * c for c in poly]
+        secret_fragments = [sum(c * pow(x, i, default_ec.n) for i, c in enumerate(poly))
```
mariano54: So, T players here can cooperate to find all of the polynomial coefficients correct?

python-impl/tests.py
```diff
@@ -342,6 +345,74 @@ def test2():
...
+        # X: a list of T indices like [1, 2, 5]
+
+        # Check underlying secret key is correct
+        r = Threshold.interpolate_at_zero(X,
+                [secret_shares[x-1].value for x in X])
+        secret_cand = PrivateKey(r)
+        assert secret_cand == master_privkey
+
+        # Check signatures
+        signature_shares = [secret_shares[x-1].sign_threshold(msg, x, X)
```
mariano54: Weren't we going to do the Lagrange interpolation at aggregation time?

src/privatekey.cpp
```diff
@@ -68,6 +68,14 @@ PrivateKey PrivateKey::FromBytes(const uint8_t* bytes, bool modOrder) {
     return k;
 }
 
+PrivateKey PrivateKey::FromBN(bn_t sk) {
+    PrivateKey k;
+    k.AllocateKeyData();
+    bn_copy(*k.keydata, sk);
+    bn_free(sk);
```
mariano54: We shouldn't free the bn since we didn't create it here, we should free it outside.

python-impl/tests.py
```diff
@@ -342,6 +345,74 @@ def test2():
...
+
+    # Step 1 : PrivateKey.new_threshold
+    for player in range(N):
+        secret_key, commi, frags = PrivateKey.new_threshold(T, N)
+        for target, frag in enumerate(frags):
+            fragments[target][player] = frag
+        commitments.append(commi)
+        secrets.append(secret_key)
+
+    # Step 2 : Threshold.verify_secret_fragment
```
mariano54: OK, so the rogue pk attack is impossible since each player verifies that the coefficients correspond to the commitments...

python-impl/tests.py
mariano54: Oh I see it's implemented in c++. i think we should have it for python too

src/test.cpp
```diff
@@ -1407,6 +1410,118 @@ TEST_CASE("AggregationInfo") {
...
+
+        // 4b. Alternatively, players may sign the message blindly, creating
+        // a unit signature share: sigShare = secretShare.SignInsecure(...)
+        // These signatures may be combined with lagrange coefficients to
+        // sign the message: signature = Threshold::AggregateUnitSigs(...)
+        // The advantage to this approach is that each player does not need
+        // to know the final list of signatories.
+
+        // For example, players 1 and 3 sign.
+        InsecureSignature sigShareU1 = secretShare1.SignInsecure(
```
mariano54: I think it's kind of strange that we're using insecure signature, but it's ok because of the verify step. Maybe we should mention this somewhere?

python-impl/keys.py
mariano54: If that's the case, we should warn people to never use their existing keys as coefficients/commitments

AWice: Not sure what you mean: the polynomial coefficients are randomly chosen.

python-impl/tests.py
AWice: done

src/test.cpp
AWice: done

src/privatekey.cpp
AWice: done

Issue #31, opened 2018-09-21
Python bindings not working on debian strech, python3


mariano54:
python3 tests.py fails with RuntimeError: Caught an unknown exception!

Issue #54, opened 2019-02-08
Potential bug in Signature deserialization and aggregation using Python


ycryptx:
```
_seed = bytes([0, 50, 6, 244, 24, 199, 1, 25, 52, 88, 192,
                19, 18, 12, 89, 6, 220, 18, 102, 58, 209,
                82, 12, 62, 89, 110, 182, 9, 44, 20, 254, 22])

msg = b'NSAWatches'
privates = [PrivateKey.from_seed(_seed).serialize() for i in range(5)]
sigs = [PrivateKey.from_bytes(i).sign(msg).serialize() for i in privates]
_sigs = [Signature.from_bytes(i) for i in sigs]
assert Signature.aggregate(_sigs), "failing on aggregate"
```

Produces a runtime error, unknown exception

mhchia:
After tracing and testing the cpp code, I think the exception is from https://github.com/Chia-Network/bls-signatures/blob/802bc2b2d7be33a309ece87ee0c32f547a2ad8d7/src/signature.cpp#L345-L347
This is because that a signature lacks the `AggregationInfo`. It can be simply solved by setting all signatures with their corresponding `message_hash` and `public_key`, before calling `Signature.aggregate`.

However, in my use case, I have a function `aggregate_signatures(Sequence[bytes]: signatures)` whose inputs are only raw bytes of the signatures, without any knowledge of message hashes and public keys. I can work around with this RuntimeError by exposing `Signature::AggregateSigsSimple` to the Python side, when calling `Signature.aggregate` with signatures having no aggregation info. I haven't understood the code inside `Signature::Aggregate` enough to determine whether exposing `AggregateSigsSimple` is a bad idea. It will be really helpful if anyone can direct me on this.

Another dumb question, is the `std::string str(b)` necessary in https://github.com/Chia-Network/bls-signatures/pull/57/files#diff-550e7950330ae6cb2a4a407140895bc9R143?

mhchia:
cc @nmarley @mariano54 Will really appreciate having you guys' input on this issue and my questions.

mariano54:
[...]

PR #56, opened 2019-03-29
Fix invalid access to data in rvalues when converting py::bytes

codablock:
Conversion like "std::string(b)" result in a temporary r value with a
short/undefined lifetime. Accessing data of it via pointers into internal
data gives undefined behavior.

The fix is to use a temporary variable which has a well defined lifetime.

The issue that is fixed here is that even when "from_xxx" methods were
called with the same data, the resulting keys were different each time.
I assume this was not noticed before as it only appear to happen with
some compiler versions.


codablock:
FYI, this fixes #31 and probably #54

PR #59, opened 2019-05-15
Use correct aggSig value for test vector

nmarley:
Hi! I am making my own implementation of this and just noticed that the current (previous to this PR) value that's showing here for signature aggregation is actually the `sig2` value from the test vectors. The aggregated sig value that should be looks more like `0x0a638..bc658` (this is in the Python code and matches my own implementation).

I have some other small adjustments to make but plan to save those for later all at once, but this one I felt was important to be changed sooner rather than later because as-is it will confuse anyone trying to implement the spec. There may be other adjustments that need to be made as well, I've only gotten to this first test as far as signature aggregation goes. Thanks!

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -283,7 +283,7 @@ since not all 32 byte sequences are valid BLS private keys
 
 ### Aggregation
 * aggregate([sig1, sig2])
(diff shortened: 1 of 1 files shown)


mariano54:
Thank you! this is also in the cpp implementation, https://github.com/Chia-Network/bls-signatures/blob/master/src/test.cpp#L69.

PR #70, opened 2019-07-24
Rename function same as C++.

mogisawa:
JS codes may be readable clearly for C++ users. Not bug fix.


mogisawa:
Sorry, i don't know i did. I'll close it.

PR #72, opened 2019-08-14
support mcl

herumi:
Hi,
I made a patch to use [mcl](https://github.com/herumi/) for some g1/g2/pairing operations instead of relic.
It gains about 20% speed for Signing and Verification on Intel Core i7-7700.

* Sigining : 1.218 ms (original) -> 0.99 ms (mcl)
* Verification : 2.24 ms (orignal) -> 1.799 ms (mcl)

I'm not good at cmake, so I added src/fp.cpp and src/bn_c384_256.cpp directly but you can link mcl/libmcl.a instead of them.

Could you try it?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "contrib/pybind11"]
 	path = contrib/pybind11
 	url = https://github.com/pybind/pybind11.git
+[submodule "contrib/mcl"]
+	path = contrib/mcl
+	url = https://github.com/herumi/mcl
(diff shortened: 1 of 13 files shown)


mariano54:
Thanks for the PR @herumi , that's a large speed up.
Right now I rather not add another dependency, since it's consensus critical code. Maybe in the future we can use MCL (and remove relic) if we really need that performance boost.

herumi:
I see, thank you for considering.

PR #75, opened 2019-08-29
Add alternative pure js implementation.

paulmillr:
Would be useful for everyone limited by c++ compilation!


mariano54:
Thank you, but it's not the same implementation and spec, so I rather not include it.

PR #78, opened 2019-09-10
Wasm port improvements

### LINKED ISSUE #76: Browser compatibility for NPM package
There is an issue with the current way Emscripten is used such that prevents this library from working in browsers.
Essentially any synchronous work on main thread in browser is a terrible idea to the point that browsers actively fight it.
With this library running in something like Chromium you'll get this warning (should have been an error actually):
> failed to compile wasm module: RangeError: WebAssembly.Compile is disallowed on the main thread, if the buffer size is larger than 4KB. Use WebAssembly.compile, or compile on a worker thread.

This is because library embeds WASM binary as base64 string. One workaround is to use `WebAssembly.instantiateStreaming(fetch('base64 content'))` and the other would be to just have a separate WASM file. Second is definitely supported by Emscripten, not sure the first one is.

This issue (since it is a warning) also leads to this:
```
Uncaught (in promise) TypeError: Cannot read property 'fromSeed' of undefined
    at Object.generateBLSKeys (bundle.js:257)
    at Wallet.createKeyPair (bundle.js:2804)
    at Node.getOrCreateAddress (bundle.js:2200)
    at async run (bundle.js:2062)
```

Because library is instantiated (kind of), but there is nothing there, since WASM compilation never happened.

nazar-pc:
Commit messages should explain what happened on each step, but here is a gist:
* Wasm is a separate file now and is not bundled inside of JavaScript
* WebAssembly instantiation is now asynchronous and works in modern browsers (see #76)
* Above causes major API change
* Compilation was tweaked to produce smaller output
* JS build script was tweaked to work on any machine with Emscripten installed regardless of specific location
* Minor tweaks to package.json and version bump

Fixes #76

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ dist
 python-impl/__pycache__/
 blspy.*.so
 
+js_build
+node_modules
+
 main
 .o
 obj/
@@ -25,6 +28,7 @@ contrib/relic/test/CTestTestfile.cmake
 contrib/gmp-6.1.2/
 contrib/libsodium-1.0.16/
 
+.idea
 .vscode
 blstest
 blstest.*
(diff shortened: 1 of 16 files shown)


mariano54:
Looks good, thank you for the contribution!

PR #115, opened 2020-06-10
Fix OpenBSD compilation issues

n1000:
* Include stdlib.h for alloca()

* Prepend relic contrib include paths, rather than append

     * Avoids conflict with "relic.h" implemented by
       "Relic - the NDBM-compatible API of QDBM", which is used by
       mutt


mariano54:
Thanks for your contribution!

PR #116, opened 2020-06-12
Include stdlib.h instead of alloca.h on FreeBSD

n1000:
Tested on FreeBSD amd64 11.3 and 12.1.


hoffmang9:
I want to double check on any upstreaming implications tomorrow and will then advise or merge this

PR #117, opened 2020-06-21
Added option to disable bindings

mxaddict:
This is for when you only want to build for c/c++ linking

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,9 +81,11 @@ set(PP_METHD "LAZYR;OATEP" CACHE STRING "")
 add_subdirectory(contrib/relic)
 add_subdirectory(src)
 
-if (EMSCRIPTEN)
-  add_subdirectory(js-bindings)
-else()
-  # emscripten can't build python bindings, it produces only javascript
-  add_subdirectory(python-bindings)
+if (!DISABLE_BINDINGS)
+  if (EMSCRIPTEN)
+    add_subdirectory(js-bindings)
+  else()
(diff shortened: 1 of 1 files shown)


hoffmang9:
The problem with your method is that the primary use of this library is to build python wheels. I'd suggest that the better path would be to add an environment variable check like C_ONLY, at that to the documentation, and have CMake check for that.

mxaddict:
> The problem with your method is that the primary use of this library is to build python wheels. I'd suggest that the better path would be to add an environment variable check like C_ONLY, at that to the documentation, and have CMake check for that.

I think I understand what you mean, but I'm not familiar with CMake syntax for checking env vars.

I'll have to do some googling and re-create this PR later 😄

mxaddict:
> The problem with your method is that the primary use of this library is to build python wheels. I'd suggest that the better path would be to add an environment variable check like C_ONLY, at that to the documentation, and have CMake check for that.

BTW, quick question, I noticed other code in contribs are git subtrees but pybind is a submodule any particular reason for this?

hoffmang9:
Yes. Unfortunately we had to mangle relic and catch is fairly static. We use pybind in bone stock upstream. Relic is moving closer to our mangling though so we'll be trying to upstream it in the next release which is real soon now. Pybind can be pulled in as a submodule or CMake will download it if you're not using git (think compiling this library under a python install in Rasberry Pi.)

hoffmang9:
See - https://cmake.org/cmake/help/latest/variable/ENV.html

mxaddict:
I'll close this for now, will re-open once I've got the time.

PR #119, opened 2020-06-29
Ietf

mariano54:
Making this pull request so I can see all the changes and start reviewing.


src/privatekey.cpp
```diff
@@ -65,34 +67,109 @@ PrivateKey PrivateKey::FromBytes(const uint8_t* bytes, bool modOrder) {
...
+
+G2Element &operator*=(G2Element &a, PrivateKey &k)
+{
+    g2_mul(a.q, a.q, *(k.keydata));
+    return a;
+}
+
+G2Element &operator*=(PrivateKey &k, G2Element &a)
+{
+    a *= k;
```
mariano54, line 146: Does this modify the input point? Isn't that bad?

src/privatekey.cpp
```diff
@@ -65,34 +67,109 @@ PrivateKey PrivateKey::FromBytes(const uint8_t* bytes, bool modOrder) {
...
+    return a;
+}
+
+G2Element operator*(G2Element &a, PrivateKey &k)
+{
+    g2_t ans;
+    g2_new(ans);
+    g2_mul(ans, a.q, *(k.keydata));
```
mariano54, line 154: We might need sec alloc instead of putting the result on the stack?

src/privatekey.cpp
```diff
@@ -65,34 +67,109 @@ PrivateKey PrivateKey::FromBytes(const uint8_t* bytes, bool modOrder) {
...
+    g2_mul(*q, base, *keydata);
+
+    const G2Element ret = G2Element::FromNative(q);
+    Util::SecFree(*q);
+    return ret;
 }
 
-PublicKey PrivateKey::GetPublicKey() const {
+/*
+PublicKey PrivateKey::GetPublicKey() const
```
mariano54, line 171: Remove unused functions

src/privatekey.hpp
```diff
@@ -51,24 +55,55 @@ friend class Threshold;
...
+    friend G2Element &operator*=(G2Element &a, PrivateKey &k);
+    friend G2Element &operator*=(PrivateKey &k, G2Element &a);
+    friend G2Element operator*(G2Element &a, PrivateKey &k);
+    friend G2Element operator*(PrivateKey &k, G2Element &a);
+
     // Serialize the key into bytes
     void Serialize(uint8_t* buffer) const;
     std::vector<uint8_t> Serialize() const;
 
+    G2Element SignG2(
```
mariano54, line 93: Need signG1 messages too

src/schemes.cpp
```diff
@@ -0,0 +1,555 @@
...
+
+using std::string;
+using std::vector;
+
+namespace bls {
+/* These are all for the min-pubkey-size variant.
+   TODO : analogs for min-signature-size
+*/
+
+const uint8_t *BasicScheme::CIPHERSUITE_ID =
```
mariano54, line 33: Can you put a comment on each one? And also a link to the spec
https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02

src/schemes.hpp
```diff
@@ -0,0 +1,293 @@
...
+namespace bls {
+
+class Core {
+    friend class BasicScheme;
+    friend class AugScheme;
+    friend class PopScheme;
+
+public:
+    // static PrivateKey KeyGen(... ikm);  TODO
+    static vector<uint8_t> SkToPk(const PrivateKey &seckey);
```
mariano54, line 44: This depends on the scheme. Maybe rename? Also why is this not in the PrivateKey class instead?

src/schemes.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+    static G1Element SkToG1(const PrivateKey &seckey);
+
+    static vector<uint8_t> Sign(
```
mariano54, line 48: I think we should remove this, we can always call G2Element.serialize() right? What's the point of adding the extra method?

src/elements.cpp
```diff
@@ -0,0 +1,478 @@
...
+    bn_new(ordMinus1);
+    g2_get_ord(ordMinus1);
+    bn_sub_dig(ordMinus1, ordMinus1, 1);
+    g2_mul(ans.q, this->q, ordMinus1);
+    return ans;
+}
+
+G2Element G2Element::FromMessage(
+    const std::vector<uint8_t>& message,
+    const uint8_t* dst,
```
mariano54, line 305: What is the user supposed to pass in here for dst? This is confusing

src/test-bench.cpp
```diff
@@ -188,12 +189,13 @@ void benchDegenerateTree() {
...
+*/
 
 int main(int argc, char* argv[]) {
-    benchSigs();
-    benchVerification();
-    benchBatchVerification();
-    benchAggregateSigsSecure();
-    benchAggregateSigsSimple();
-    benchDegenerateTree();
+    // benchSigs();
```
mariano54, line 195: Uncomment

src/test.cpp
```diff
@@ -14,26 +14,30 @@
...
+    SECTION("Test vectors 1")
+    {
         uint8_t seed1[5] = {1, 2, 3, 4, 5};
         uint8_t seed2[6] = {1, 2, 3, 4, 5, 6};
         uint8_t message1[3] = {7, 8, 9};
-        
+
         PrivateKey sk1 = PrivateKey::FromSeed(seed1, sizeof(seed1));
         PublicKey pk1 = sk1.GetPublicKey();
         Signature sig1 = sk1.Sign(message1, sizeof(message1));
```
mariano54, line 43: Uncomment and fix tests

PR #125, opened 2020-07-29
Fix readme error

Raycoms:
Readme uses Signature::Aggregate to aggregated Public Keys, this does not work. It has to be PublicKey::Aggregate


mariano54:
Thanks. In the lastest version, the whole API is changed, so we will do a revamp of the whole readme.

PR #129, opened 2020-08-13
Ms.fix bugs2

mariano54:
- Ensure subgroup membership check, in compliance with spec
- Use different DST for signature vs proof of possession
- Make elements immutable to prevent security bugs
- Remove unused/mutable methods
- Make class members private when they don't need to be public
- Perform pairings 250 at a time to not overflow stack
- Python implementation complete, and matches c++


python-impl/ec.py
```diff
@@ -511,6 +515,52 @@ def twist(point: AffinePoint, ec=default_ec_twist) -> AffinePoint:
...
+    mapvals[3] *= pow(z, 3)
+
+    Z = mapvals[1] * mapvals[3]
+    X = mapvals[0] * mapvals[3] * Z
+    Y = mapvals[2] * mapvals[1] * Z * Z
+    return JacobianPoint(X, Y, Z, P.infinity, ec)
+
+
 """
 Copyright 2018 Chia Network Inc
```
hoffmang9, line 572: Worth bumping this

python-impl/impl-test.py
```diff
@@ -4,6 +4,7 @@
 
 from fields import Fq, Fq2, Fq6, Fq12
 from ec import (
+    JacobianPoint,
```
hoffmang9, line 7: We should add these tests to the github actions so that we're testing the python implementation on every build too. Happy to do that for you if you'd like.

mariano54: It takes like 10 minutes to run the python implementation tests, do you think it is worth it?

hoffmang9: An option would be to make it an "on click" action. It's there on every run but doesn't run unless you go over to actions and click Run. Thoughts?

mariano54:
Yeah we could do that.

PR #136, opened 2020-10-20
Add fast pub aggregation

Raycoms:
At the moment public key aggregation is only possible through the "+" operator override. This is extremely slow for large quantities of public keys.

This pull requests adds a dedicated aggregation function for public keys.
The speedup factor running a smaller benchmark already resulted in around 25x

---- Old------
PopScheme verification
Total: 500 runs in 74 ms
Avg: 0.148 ms

---- New------
PopScheme verification
Total: 500 runs in 3 ms
Avg: 0.006 ms


mariano54 (commented):
Thank you for the changes. Do you mind making the same change for G2?

PR #139, opened 2020-11-04
Leakwjb

wjblanke:
Fix sodium_malloc leak in BNWrapper. Hopefully will fix farmer crash due to sodium_malloc fails.


hoffmang9 (approved):
Looks good to me

PR #140, opened 2020-11-04
Allow validation of infinity sig

mariano54:


mariano54:
Actually I think this is not compliant with IETF spec (no pks means that signature validation should fail), if this is the case, I will not merge

hoffmang9:
@bramcohen is getting what eth-2 is doing on this so we should have that shortly.

hoffmang9:
Closing - used https://github.com/Chia-Network/bls-signatures/pull/149 instead

PR #141, opened 2020-11-04
fix self-assignment on BNWrapper

arvidn:
as well as move construction and move assignment. I bumped C++ version to 17 for `std::exchange()`. Is there a reason to stay on C++11 in this project?


hoffmang9 (approved):
Looks good

PR #142, opened 2020-11-05
yet another leak

wjblanke:


hoffmang9 (approved):
Looks good

src/elements.cpp
```diff
@@ -127,6 +129,7 @@ void G1Element::CheckValid() const {
     } catch (...) {
         throw std::invalid_argument("Relic reports invalid argument given");
     }
+    bn_free(order);
```
arvidn, line 132: technically this will leak in any of the error cases. I think this is probably good enough for a short-term fix. long term I think it would be a lot more reliable to wrap the `g1_t` and `bn_t` types in types that manage the resources.

src/elements.cpp
```diff
@@ -388,8 +398,11 @@ G2Element operator*(const G2Element& a, const bn_t& k)
     bn_new(nonConstK[0]);
     bn_copy(nonConstK[0], k);
     g2_mul(ans, nonConstA.q, nonConstK[0]);
+    bn_free(nonConstK[0]);
```
arvidn, line 401: my understanding of the relic types, `g1_t`, `ep_t` etc. is that the are pointers to `calloc()`ed memory (allocated by their `*_new()` functions). Allocating the `bn_t` object *itself* in secure memory is most likely pointless, since the *value* is stored on the regular heap.

see [this](https://github.com/relic-toolkit/relic/blob/50f0b836cbf93bc255f41ab730463a38330114b9/include/relic_bn.h#L142).

src/hdkeys.hpp
```diff
@@ -97,6 +97,7 @@ class HDKeys {
         Util::SecFree(ikmHkdf);
         Util::SecFree(skBn);
         Util::SecFree(okmHkdf);
+        Util::SecFree(skBytes);
```
arvidn, line 100: this leak could possibly cause the exhaustion of libsodium allocations. But the security of this doesn't seem all the way thought through. Apart from the fact that allocating the `bn_t` object itself in secure memory, the private key is then copied to the stack.

arvidn (approved):

hoffmang9 (approved):
looks good to me!

arvidn (approved):

PR #144, opened 2020-11-06
fix resource management in PrivateKey class

arvidn:
add assignment operators to properly copy value (instead of pointer) or move (instead of copying pointer, resulting in double-free)


hoffmang9 (approved):
lgtm

PR #145, opened 2020-11-15
Default to dynamically linking libsodium, static link on wheels

hoffmang9:
- Now, if libsodium is found the build process will default to dynamically linking libsodium. This should ease cross platform building.
- If $CIBUILDWHEEL is set in the environment, libsodium will be statically linked. That is the default when building binary wheels on ci.
- Moved valgrind to test the statically linked libsodium version of bls-library.
- "By hand" (using strings) validated the published wheels for all four targets are getting statically linked from cibuildwheel.
- Have relic upstream a fix to FreeBSD and OpenBSD
-Correct "DSEED" typo - though note that "SEED" defaults to the correct random device so there should be no historical issue.

### DIFF
--- a/.github/workflows/build-aarch64.yml
+++ b/.github/workflows/build-aarch64.yml
@@ -61,10 +61,9 @@ jobs:
         python -m cibuildwheel --output-dir dist
       env:
         # build python 3.7 and 3.8
-        CIBW_ENVIRONMENT: "BLSALLOC_SODIUM=1 SODIUM_STATIC=1"
         CIBW_BUILD: cp37-* cp38-*
         CIBW_MANYLINUX_AARCH64_IMAGE: manylinux2014
-        CIBW_BUILD_VERBOSITY_LINUX: 1
+        CIBW_BUILD_VERBOSITY_LINUX: 0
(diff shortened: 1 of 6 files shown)


hoffmang9:
I have also confirmed that the bsd compile issues are either a relic issue or we're note handing relic the right options. Debugging that now. See these earlier fixes from @n1000 :
https://github.com/Chia-Network/bls-signatures/commit/5a7c70385a9693eba442604719cd167d6fc56745
https://github.com/Chia-Network/bls-signatures/commit/fcc10ef18106017b506c4aa29a3624558eefc8dc

hoffmang9:
I've made an upstream pull request to Relic. - https://github.com/relic-toolkit/relic/pull/164

Once that's complete we will probably need to address the changes previously made to our Cmakelist.txts outlined here - https://github.com/Chia-Network/bls-signatures/commit/5a7c70385a9693eba442604719cd167d6fc56745

hoffmang9:
image
With the Relic upstream, `pip install .` is working fine on FreeBSD!

wjblanke (approved):
Does SODIUM_STATIC need to be based on CI_BUILDWHEEL as well?

Approving but we may need to revisit

PR #146, opened 2020-11-17
Versioninfowjb

wjblanke:
Correctly quote version define under python 3.9


hoffmang9 (approved):
Looks very good to me

PR #147, opened 2020-11-17
src: Throw exceptions instead of std::string, drop some redundant code

xdustinface:
See commit messages

### DIFF
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -124,12 +124,8 @@ void G1Element::CheckValid() const {
     ep_set_infty(unity);
     if (g1_cmp(point, unity) != RLC_EQ)
         throw std::invalid_argument("Given G1 element failed in_subgroup check");
-    try {
-        BLS::CheckRelicErrorsInvalidArgument();
-    } catch (...) {
-        throw std::invalid_argument("Relic reports invalid argument given");
-    }
     bn_free(order);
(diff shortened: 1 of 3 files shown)


mariano54:
Looks good to me

hoffmang9 (approved):
lgtm

hoffmang9:
Thanks for the PR!

src/elements.cpp
```diff
@@ -124,12 +124,8 @@ void G1Element::CheckValid() const {
     ep_set_infty(unity);
     if (g1_cmp(point, unity) != RLC_EQ)
         throw std::invalid_argument("Given G1 element failed in_subgroup check");
-    try {
-        BLS::CheckRelicErrorsInvalidArgument();
-    } catch (...) {
-        throw std::invalid_argument("Relic reports invalid argument given");
-    }
     bn_free(order);
+    BLS::CheckRelicErrors();
```
nmarley, line 128: @xdustinface What is the reason that you called this underneath `bn_free` instead of above it like the previous try/catch block?

xdustinface: The idea was to ensure `order` gets even freed i.e.  `bn_free` gets called even if `CheckRelicErrors` throws. But seems like it should be moved up 2 more lines to absolutely ensure that.. 🤔

PR #158, opened 2021-02-03
schemes: Some improvements/simplifications

xdustinface:
Avoid some element creation and copying for no reason, be a bit more descriptive and type correct for some variables, drop some redundant code and simplify + improve readability in some places. Just walk through the commits hierarchically for easier review. Le me know your thoughts or concerns :)

Based on #150.

### DIFF
--- a/src/schemes.cpp
+++ b/src/schemes.cpp
@@ -258,10 +258,7 @@ bool BasicSchemeMPL::AggregateVerify(const vector<G1Element> &pubkeys,
 
 G2Element AugSchemeMPL::Sign(const PrivateKey &seckey, const vector<uint8_t> &message)
 {
-    vector<uint8_t> augMessage = seckey.GetG1Element().Serialize();
-    augMessage.reserve(augMessage.size() + message.size());
-    augMessage.insert(augMessage.end(), message.begin(), message.end());
-    return CoreMPL::Sign(seckey, augMessage);
+    return Sign(seckey, message, seckey.GetG1Element());
 }
 
 // Used for prepending different augMessage
--- a/src/schemes.cpp
+++ b/src/schemes.cpp
@@ -31,33 +31,17 @@ enum InvariantResult { BAD=false, GOOD=true, CONTINUE };
 
 // Enforce argument invariants for Agg Sig Verification
 InvariantResult VerifyAggregateSignatureArguments(
-    const vector<G1Element> &pubkeys,
-    const vector<vector<uint8_t>> &messages,  // unhashed
+    const size_t nPubKeys,
+    const size_t nMessages,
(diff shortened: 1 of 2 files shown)


src/schemes.cpp
```diff
@@ -171,15 +155,15 @@ bool CoreMPL::AggregateVerify(
     const vector<uint8_t> &signature)
 {
     int n = pubkeys.size();
+    G2Element signatureElement = G2Element::FromByteVector(signature);
     if (n <= 0 || n != messages.size()) {
```
mariano54: It looks like we are only checking args if we go into this if statement. Why not always check, for consistency with the other call locations?

I think a better way to do it, is to not check in the caller location. And instead do this:

if (!VerifyAggregateSignatureArguments(n, messages.size(), signatureElement) {
    return False
}

src/schemes.cpp
```diff
@@ -393,14 +393,15 @@ bool PopSchemeMPL::FastAggregateVerify(
     const vector<uint8_t> &message,
     const G2Element &signature)
 {
+    if (pubkeys.size() <= 0) return false;
```
mariano54: Actually the infinity element signature should validate if there are no messages or public keys. I think we forgot to handle this edge case for FastAggregateVerify

mariano54 (commented):
These are good changes, thanks! I'd like to hear your thoughts on my two comments

src/schemes.cpp
mariano54: We don't have to address this is this PR. But the behavior should be the same for all schemes when verifying with no keys/message I think

xdustinface: Hm, actually it shouldn't matter in this case if i got it right. You will always have a message here which means https://github.com/Chia-Network/bls-signatures/blob/5d5da2778916cf39e5908372a36fd20c14f88724/src/schemes.cpp#L39 will anyway return `BAD` aka `false` and it couldn't even run into the "allow infinity" case.

src/schemes.cpp
xdustinface: Makes sense 👍 I just unified it in all places in 0de8f04e19c2ce03761c1bbd1b75f44a30049e19, thoughts?

xdustinface:
> I'd like to hear your thoughts on my two comments

Replied :) and also added one more: 316fb91

src/schemes.cpp
mariano54: Oh okay that makes sense. These methods take in one public key and one message so there is no need to check

mariano54:
LGTM

mariano54 (approved):

hoffmang9:
Will merge shortly

PR #170, opened 2021-02-24
src|python: Drop `G1Element::Infinity()` and `G2Element::Infinity()`

xdustinface:
They are Infinity by default


mariano54 (approved):
I avoided doing this because it's not clear whether it's generator or infinity. Also note to chia developers, we need to change uses of infinity in chia-blockchain after updating blspy. I'm fine with merging this though.

PR #182, opened 2021-03-08
git|contrib: Drop relic submodule

xdustinface:
This is just confusing because its the `Chia-Network` fork and its not really related to the relic version that cmake pulls.


hoffmang9 (approved):
Thanks - hadn't had time...

PR #183, opened 2021-03-09
Confirm libsodium, gmp linking in wheels

hoffmang9:


hoffmang9:
Deleting as CI correctness confirmed.

PR #187, opened 2021-03-17
Update to RELIC's change from COMP to CFLAGS.

dfaranha:
Quick update to reflect that RELIC changed naming of a variable to be closer to CMake conventions.


hoffmang9:
Note that the ARM64 issue is not you but a strange git holdover issue on my todo list to fix.

hoffmang9 (approved):
Looks good.

PR #188, opened 2021-03-17
fix typo in test-bench

arvidn:


hoffmang9:
There is git issue that I need to track down on ARM64. I'm aware and hopefully can fix it shortly.

PR #189, opened 2021-03-22
use int instead of double for counting test iterations

arvidn:
I don't think there's a good reason to use `double` to count iterations. It's inherently an integer.


hoffmang9 (approved):
+1

PR #196, opened 2021-03-31
remove redundant validation checks

arvidn:
`ep_map_dst()` and `ep2_map_dst()` ought to return valid points, the validation of their output is redundant (and expensive).

These are my timings before and after this change:

```
Batch verification
Total: 100000 runs in 133637 ms
Avg: 1.33637 ms
```

```
Batch verification
Total: 100000 runs in 73739 ms
Avg: 0.73739 ms
```

### DIFF
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -88,7 +88,7 @@ G1Element G1Element::FromMessage(const Bytes& message,
 {
     G1Element ans;
     ep_map_dst(ans.p, message.begin(), (int)message.size(), dst, dst_len);
-    ans.CheckValid();
+    assert(ans.IsValid());
     return ans;
 }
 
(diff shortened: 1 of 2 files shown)


arvidn:
@xdustinface is the premise for this change correct?

mariano54 (approved):

src/elements.cpp
```diff
@@ -100,24 +100,27 @@ G1Element G1Element::Generator()
...
     // check if inside subgroup
     bn_t order;
     bn_new(order);
     g1_get_ord(order);
 
-    G1Element point = *this * order;
-    if (point != G1Element())
-        throw std::invalid_argument("Given G1 element failed in_subgroup check");
+    const G1Element point = *this * order;
     bn_free(order);
```
arvidn, line 117: as a bonus, `order` would leak if the check failed previously.

wjblanke (approved):
looks ok, as long as they are good!

xdustinface (changes requested):
> @xdustinface is the premise for this change correct?

I think it makes sense 👍 But i have two points:

- If this will be merged there should probably still be a `BLS::CheckRelicErrors()` call where you removed the subgroup checks.
- I looked into the related relic parts a bit and im not sure how to interpret and weight [this comment](https://github.com/relic-toolkit/relic/blob/main/src/ep/relic_ep_map.c#L155-L159). Can it fail or not? 🤔 And if it fails, does it then silently end up in the wrong group (which would be a breaker for this PR) or will it throw? I guess this should be answered first but i probably can't with certainty 🙈

arvidn:
> If this will be merged there should probably still be a `BLS::CheckRelicErrors()` call where you removed the subgroup checks.

I agree. I added those checks back.

> I looked into the related relic parts a bit and im not sure how to interpret and weight [this comment](https://github.com/relic-toolkit/relic/blob/main/src/ep/relic_ep_map.c#L155-L159). Can it fail or not? :thinking: And if it fails, does it then silently end up in the wrong group (which would be a breaker for this PR) or will it throw? I guess this should be answered first but i probably can't with certainty :see_no_evil:

@dfaranha or @kwantam could you comment on the assumption that `ep_map_dst()` and `ep2_map_dst()` always either:

* return valid points in the G1 and G2 subgroups respectively
* fail and set the error code in `core_get()->code`.

dfaranha:
From @kwantam

arvidn:
> PS: I have implemented faster subgroup checks (in a80e1db252c78854b1017a8fc97e6b962502031d) as per (https://eprint.iacr.org/2019/814.pdf). Does the benchmarking here reflect this? Not sure about which version was measured.

I noticed that yesterday. This benchmark is running against an older version of relic, and we do our own subgroup check as well.
I have a patch ready locally that bumps the relic dependency and removes our own subgroup check.

PR #198, opened 2021-03-31
relic's main branch is "main"

arvidn:
It seems it used to be `master`.

`relic_test.h` now defines a macro called `TEST_CASE` which conflicts with catch's macro with the same name. It doesn't seem like we need anything out of that header though.

### DIFF
--- a/.github/workflows/relic-nightly.yml
+++ b/.github/workflows/relic-nightly.yml
@@ -24,12 +24,12 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


arvidn:
the nightly build against relic is not triggered by pull requests, but run on a schedule. it started failing about 15 days ago.

https://github.com/Chia-Network/bls-signatures/runs/2236289722?check_suite_focus=true

hoffmang9 (approved):
There was another upstream change requested by the Dash folks so I thought that was why it had begun failing...

PR #199, opened 2021-04-01
releases the Python GIL for methods used in loading plots

wbendick:


mariano54 (changes requested):
Can you explain why you chose these methods and not all methods, particularly why the pairing method was not chosen (which is the slowest).
Also can you fix the indentation? 
Finally, it would be useful to get performance numbers for the new changes (how long does it take to load 1000 plots compared to before, etc).

python-bindings/pythonbindings.cpp
```diff
@@ -26,6 +26,14 @@ namespace py = pybind11;
 using namespace bls;
 using std::vector;
 
+class PythonGIL {
```
arvidn, line 29: pybind11 already has a class that (I believe) does this: `pybind11::gil_scoped_release`

python-bindings/pythonbindings.cpp
```diff
@@ -362,6 +374,7 @@ PYBIND11_MODULE(blspy, m)
         .def(
             "__bytes__",
             [](const G1Element &ele) {
+	        PythonGIL release_lock;
                 vector<uint8_t> out = ele.Serialize();
                 py::bytes ans = py::bytes(
```
arvidn, line 379: you have to acquire the lock again before constructing `py::bytes`.

python-bindings/pythonbindings.cpp
```diff
@@ -362,6 +374,7 @@ PYBIND11_MODULE(blspy, m)
         .def(
             "__bytes__",
             [](const G1Element &ele) {
+	        PythonGIL release_lock;
```
arvidn, line 377:
```suggestion
                vector<uint8_t> out = [&] {
	            PythonGIL release_lock;
	            return ele.Serialize();
	        };
```

arvidn (commented):
If there's a good reason to use your own GIL release type, I think there should be a comment explaining it.

I agree that the most expensive functions should also receive this treatment.

There's also a risk that if you do this on very cheap operations, you mostly just slow them down.

wbendick:
> Can you explain why you chose these methods and not all methods, particularly why the pairing method was not chosen (which is the slowest).

I was only speeding up the code that affects harvester startup because that's what I'm testing right now. I think the library should do this in all places where possible in the future though.

> Also can you fix the indentation?

Done.

> Finally, it would be useful to get performance numbers for the new changes (how long does it take to load 1000 plots compared to before, etc).

Loading several thousand plots on one machine using this improvement decreased the runtime from 1274 seconds to 202 seconds.

wbendick:
> If there's a good reason to use your own GIL release type, I think there should be a comment explaining it.

This release type is considerably faster than gil_scoped_release for situations where you won't be reacquiring the GIL. Using gil_scoped_release negates most of the benefit of this optimization because the runtimes of the calls are small in an absolute sense. We just make a large number of times.

python-bindings/pythonbindings.cpp
wbendick: This release type is considerably faster than gil_scoped_release for situations where you won't be reacquiring the GIL. Using gil_scoped_release negates most of the benefit of this optimization because the runtimes of the calls are small in an absolute sense. We just make a large number of times.

python-bindings/pythonbindings.cpp
wbendick: You are correct. Fixed.

python-bindings/pythonbindings.cpp
```diff
@@ -362,7 +377,10 @@ PYBIND11_MODULE(blspy, m)
         .def(
             "__bytes__",
             [](const G1Element &ele) {
-                vector<uint8_t> out = ele.Serialize();
+                vector<uint8_t> out;
+                Py_BEGIN_ALLOW_THREADS
```
arvidn, line 381: the only way this makes sense to me is if `Py_BEGIN_ALLOW_THREADS` introduces a new scope. Does it?

mariano54 (commented):
Could you also add it for all the verify methods? This could allow signature verification optimizations in Chia blockchain

python-bindings/pythonbindings.cpp
wbendick: It does

python-bindings/pythonbindings.cpp
```diff
@@ -64,7 +77,7 @@ PYBIND11_MODULE(blspy, m)
             [](const PrivateKey &k, const py::object &memo) {
                 return PrivateKey(k);
             })
-        .def("get_g1", [](const PrivateKey &k) { return k.GetG1Element(); })
+        .def("get_g1", [](const PrivateKey &k) { PythonGIL release_lock; return k.GetG1Element(); })
```
mariano54, line 80: Indentation

mariano54 (approved):
LGTM, apart from the indentation

python-bindings/pythonbindings.cpp
wbendick: I don't see the indentation problem. It matches the surrounded expressions and uses spaces.

PR #200, opened 2021-04-01
Simplify IsValid()

arvidn:
New versions of relic includes an optimized subgroup check for G1 and G2. Not only do we not need to do our own subgroup check, but we also get a good speedup by just upgrading relic.

In `main` right now, our `IsValid()` call is dominated by the cost to `g1_is_valid()` and `g2_is_valid()` calls (>90% of the time is spent in those calls).

Here's a comparison between just the current version of relic vs. head of `main` (both with this patch to `IsValid()`):

```
Public key validation
Total: 100000 runs in 33414 ms
Avg: 0.33414 ms

Signature validation
Total: 100000 runs in 68799 ms
Avg: 0.68799 ms
```

```
Public key validation
Total: 100000 runs in 14181 ms
Avg: 0.14181 ms

Signature validation
Total: 100000 runs in 18321 ms
Avg: 0.18321 ms
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -7,7 +7,7 @@ include(FetchContent)
 if (DEFINED ENV{RELIC_MAIN})
   set(RELIC_GIT_TAG "origin/main")
 else ()
-  set(RELIC_GIT_TAG "03a7c3b7fa43c0da6f3720e341f7d4f6a6d6f21e")
+  set(RELIC_GIT_TAG "1885ae3b681c423c72b65ce1fe70910142cf941c")
 endif ()
 
 message(STATUS "Relic will be built from: ${RELIC_GIT_TAG}")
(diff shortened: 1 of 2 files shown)


arvidn:
@dfaranha if you're interested in the speedup with Sean Bowe's subgroup check ^^

hoffmang9 (approved):
Looks good to me but needs at least one more "not me" approval.

arvidn:
for reference. Fast subgroup checking is done in `g1_is_valid()` and `g2_is_valid()` since this commit: https://github.com/relic-toolkit/relic/commit/a80e1db252c78854b1017a8fc97e6b962502031d

dfaranha:
Glad to hear! I have even faster variants of these checks on the works, but first I'm trying to prove that they are sufficient. :)

hoffmang9:
@dfaranha Do you have a guesstimate of the speed up you'd obtain if they check out? We're in process of setting our smart coin virtual machine costs and having even a vague sense of what may be out there is valuable.

Also, I've set up a nightly build of this library against relic here - https://github.com/Chia-Network/bls-signatures/actions/workflows/relic-nightly.yml It also can be run on demand and if you'd like I can get you access to that button.

mariano54 (approved):

dfaranha:
If confirmed, membership testing in both G_1 and G_2 will become essentially free. I'm confident about the approach for G_1, not as much for G_2 but checking with Sean Bowe. Testing membership in G_T will continue to be expensive until we can factor the large cofactors.

PR #207, opened 2021-04-21
Fix support to latest version of RELIC

dfaranha:
Minor fixes to take into account recent changes in RELIC:
* Rename `master` branch to `main`
* Remove unnecessary include of `relic_test.h`
* Replaced recently renamed `bn_init` with something simpler (assuming that `ALLOC == AUTO` remains in place)


hoffmang9 (approved):
This got caught in launch and then pooling focus. I'll be happy to have the nightly's passing after we get this merged. Thanks again!

PR #212, opened 2021-05-07
Revive javascript bindings and port python tests to javascript.

prozacchiwawa:
The javascript bindings that were previously in this code followed an API that appears to have been withdrawn in revision c6010c8, so this PR brings them into fairly close parity with the python bindings.
A few liberties were taken with naming and capitalization to follow javascript idia more closely, and the tests in test.py were ported.
CMake files have been modified to refer to relic in its new build location as cmake presents it and a build script emsdk_build.sh has been provided which uses emsdk to build the js library and bindings, saving some complexity.
A convenience script for running the javascript tests is provided.


hoffmang9 (approved):
Thank you!

PR #214, opened 2021-05-08
Remove outdated ECC docs line

Nikolaj-K:
* Removing an outdated comment standing in conflict with the code. 
* Taking the chance of completing some sentences in the other docs there as well.

The doc suggests the standard secp256k1 is being used as default, but actually more recent curve parameters are used in the line below. 
There's actually two identical such docs being removed, one in the `ec` file and one in the `pairing` file. As far as I can see, nothing should stand in the way or importing one of them in the other (e.g. removing the one in the `pairing` file and there importing it from `ec`). 
(The module level [constants should](https://www.python.org/dev/peps/pep-0008/#constants) also be all caps, as in `DEFAULT_EC`. At least I think it's a constant.)

----
Some notes for those files unrelated to the docs. 
* I wonder if python3 could be required expectation for this repo too, in which case I wonder if the `from __future__` import is still necessary.
* I note that some `is_instance` type checks follow exceptions being thrown, but occasionally in those two files, merely `False` is returned. I think one can throw there too.
* There's probably some if-checks that can be conveniently subsumed in mere boolean operations.
* The language of the docstrings are a bit inconsistent, but okay that's a nit
* What could be a bug - I just want to bring it up - is that the two last functions in /pairing.py make use of `default_ec.` in the body despite `ec=default_ec`being passed as an argument. On first glance it feels like `ec.` is intended to be used there.

### DIFF
--- a/python-impl/ec.py
+++ b/python-impl/ec.py
@@ -11,7 +11,6 @@ from util import hash256
 # Struct for elliptic curve parameters
 EC = namedtuple("EC", "q a b gx gy g2x g2y n h x k sqrt_n3 sqrt_n3m1o2")
 
-# use secp256k1 as default
 default_ec = EC(*bls12381.parameters())
 default_ec_twist = EC(*bls12381.parameters_twist())
 
@@ -37,7 +36,7 @@ class AffinePoint:
 
     def is_on_curve(self) -> bool:
         """
-        Check that y^2 = x^3 + ax + b
+        Check that y^2 = x^3 + ax + b.
         """
         if self.infinity:
             return True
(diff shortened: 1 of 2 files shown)


python-impl/pairing.py
```diff
@@ -8,7 +8,6 @@
 # Struct for elliptic curve parameters
 EC = namedtuple("EC", "q a b gx gy g2x g2y n h x k sqrt_n3 sqrt_n3m1o2")
 
-# use secp256k1 as default
 default_ec = EC(*bls12381.parameters())
```
Nikolaj-K, line 11: This seems to be duplicated constant of line 15 in /ec.py.

python-impl/ec.py
```diff
@@ -308,7 +307,7 @@ def bytes_to_point(buffer: bytes, ec, FE) -> JacobianPoint:
 
 def y_for_x(x, ec=default_ec, FE=Fq):
     """
-    Solves y = sqrt(x^3 + ax + b) for both valid ys
```
Nikolaj-K, line 311: This line says `Solves` and the above one `Check that`. 
I think a convention (active or passing) should be chosen. [Standard python](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) would probably be to do the latter (active job description, as in "Check that" and "Solve")

mariano54 (approved):

PR #221, opened 2021-05-25
Add asserts to python binding README and test

Nikolaj-K:
No all bool checks have asserts, so here I add the others.
I also order some lines in a way that things are closest where they are actually used.


mariano54 (commented):
This makes the README a little bit harder to read, so I prefer to not include this one

Issue #225, opened 2021-06-08
Build fails when GMP_NAME is found to be a shared lib `*.so` file


madMAx43v3r:
See: https://github.com/Chia-Network/bls-signatures/blob/4e21aa6742d67562920ac67507e90b722116b14b/src/CMakeLists.txt#L56

```
[ 96%] Linking C static library ../lib/librelic_s.a
[ 96%] Built target relic_s
Scanning dependencies of target combined_custom
Scanning dependencies of target runbench
/bin/ar: /usr/lib64/libgmp.so: File format not recognized
make[2]: *** [lib/bls-signatures/src/CMakeFiles/combined_custom.dir/build.make:79: lib/bls-signatures/src/CMakeFiles/combined_custom] Error 1
make[1]: *** [CMakeFiles/Makefile2:537: lib/bls-signatures/src/CMakeFiles/combined_custom.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

xorinox:
Issue was already brought up in 2019 and then closed :(   https://github.com/Chia-Network/bls-signatures/issues/68

xorinox:
On Fedora 33 Server I installed the static versions of the libs gmp and sodium:
`sudo yum install libsodium-static.i686 gmp-static.i686`

After that I could compile bls-signatures as per readme without issues.

PR #231, opened 2021-06-14
Updates to cmake to work around missing statics

emlowe:
Setup CMAKE to require C++17 - this helps with places where the default compiler does not (Centos 7) and errors out early
Also, configure to avoid making the "combined custom" library libbls.a if the platform does not have the static versions of libgmp or libsodium. The library and test programs still work, however, because the dynamic version will be linked to the executable (ie runtest, etc)
This does mean on such a system that when using libbls.a one does need to link to the dynamic libraries, but this is preferable to not building at all and throwing strange errors. This should fix issue #225


hoffmang9 (approved):
lgtm

emlowe:
I also made a small change for 32-bit systems (which are actually pretty rare now - most common Linux distros don't support them anymore).
Tested on Centos7 32-bit
Tested on Centos7 64-bit

hoffmang9 (approved):
Nice add

hoffmang9 (approved):
AOK

PR #233, opened 2021-06-15
update glob-parent version

nirajpathak13:
Updated glob-parent vulnerability as reported by dependabot:
https://github.com/Chia-Network/bls-signatures/security/dependabot/js-bindings/package-lock.json/glob-parent/open


hoffmang9 (approved):
If it passed tests, we're good.

PR #237, opened 2021-06-28
Add functions to Python binding to support pairing cache

rostislav:


xdustinface (commented):
@rostislav Looks good but one question. Wouldn't it make sense to only partially revert 4a1013a and keep `BNWrapper` dropped or do you also need it for something?

rostislav:
@xdustinface I agree that `BNWrapper` could be dropped. I'll prepare an update for this.

rostislav:
I've removed `BNWrapper` and added some tests for the new functions.

xdustinface (changes requested):
@rostislav I just realized there is even more code we should not revert. See 7615e4ff3430d791a7aaf85f9e326b1848e2c90c.

xdustinface (approved):

PR #249, opened 2021-07-16
Add fix for RELIC parsing invalid inputs.

dfaranha:
This patch handles RELIC errors when invalid inputs are provided.


hoffmang9 (approved):
Thanks for this - will also bump relic as you mentioned.

PR #252, opened 2021-07-19
[RFC] Make credible test set in .spec files and get everything passing according to new bls api

prozacchiwawa:
This brings the ci runnable test set up to date with the current API of bls-signatures as made in the python bindings and reproduced in jsbindings.cpp.

PublicKey and Signature became G1Element and G2Element
Signature lost aggregation info and its associated methods
Aggregation related methods in Signature and PublicKey became scheme specific in BasicSchemeMPL, AugSchemeMPL and PopSchemeMPL (aggregate, aggregate_verify).
Key seed picked up the requirement of being at least 32-bytes in length and changed salt string
PublicKey and Signature picked up requirement of having the first bit set in a serialized value

I've preserved as much of the spirit of the original as possible and used existing test byte data where possible.

Karma has updated significantly in the interim and runs webpack differently, so some updating and adjustment were required for the browser side test run.


js-bindings/tests/PrivateKey.spec.js
```diff
@@ -1,110 +1,100 @@
...
+const {Buffer} = require('buffer');
 
-function getSeedAndFinferprint() {
-    return {
-        seed: Uint8Array.from([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
-        fingerprint: 0xddad59bb
-    };
-}
-
+// Values lifted from test.js, which is ported from test.py and is a more
```
aqk, line 6: Good comment, thanks.

aqk (approved):
Excited to have tests for this API

aqk:
@mariano54 can you take a look at these BLS tests?

Issue #251, opened 2021-07-18
Decompression of invalid input succeeds


guidovranken:
OSS-Fuzz issue: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36301

```cpp
#include <relic_conf.h>
#include <bls.hpp>

int main(void)
{
    const std::vector<uint8_t> compressed = {0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
    const auto g1 = bls::G1Element::FromBytes(::bls::Bytes(compressed));
    return 0;
}
```

This throws an exception at https://github.com/Chia-Network/bls-signatures/tree/7457c4bdc0991c8911f725d32459d7f0b0b730c3 but does not throw an exception at current HEAD (https://github.com/Chia-Network/bls-signatures/tree/3c3cfb40b5669b41225fba91e2b3461a38989633).

I think it should throw an exception because that's what it has always done, and aligns with the behavior of blst.

CC @dfaranha

wjblanke:
@dfaranha this appears to be the relic commit that removes the thrown exception. I found this by regression testing not sure what it happening

https://github.com/relic-toolkit/relic/commit/c7177c87118333db166175fe635b953cea8d8c58

wjblanke:
Some info from Kevin, this is an issue on our side

this commit obviously fixed a bug but we still accept this zero values to be okay since we just set the first byte of the buffer to 0x02 or 0x03 if the first byte of the input is NOT signaling its infinity although the rest of the buffer is zero.

guidovranken:
> @dfaranha this appears to be the relic commit that removes the thrown exception. I found this by regression testing not sure what it happening
> 
> [relic-toolkit/relic@c7177c8](https://github.com/relic-toolkit/relic/commit/c7177c87118333db166175fe635b953cea8d8c58)

@wjblanke : this commit is a fix for https://github.com/relic-toolkit/relic/issues/206

[...]

PR #253, opened 2021-07-20
src|cmake: Fix some `FromBytes` decoding issues, add more test cases

xdustinface:
This is a (i think the proper) fix for #251 which basically has been introduced by a bug-fix in relic, see https://github.com/relic-toolkit/relic/commit/c7177c87118333db166175fe635b953cea8d8c58 and https://github.com/relic-toolkit/relic/issues/206. Please also have a look here @dfaranha :)

a90d56a expands the test case to cover the issue detected in #251 and e9f60d1 is the actual fix.


hoffmang9 (approved):
lgtm

xdustinface:
Okay, turned out that there are more "first byte values" where it should pass because the 4 lower bits of the first byte are also used as data. I adjusted the test and fixed an issue in the zero bytes check.. now it should have the same behavior like before the relic bug-fix which introduced the change of the behavior. But at the end i guess this should also be fixed in relic itself @dfaranha?

Another thing i realized is that with this commit https://github.com/relic-toolkit/relic/commit/1620a03b388e50acd68ed9c88d7cd82ec5490ce4 we get more invalid cases, see 1fa53dc91a30b90f007f0991c8c3189dc0e9db43 which is required after that commit . I miss some background here to judge if all of the values in here 
```
0x85, 0x86, 0x87, 0x88, 0x89, 0x8C, 0x8F, 0x91, 0x93, 0x94, 0x96, 0x98, 0x99, 0x9A, 0xA5, 0xA6, 0xA7, 0xA8, 0xA9, 0xAC, 0xAF, 0xB1, 0xB3, 0xB4, 0xB6, 0xB8, 0xB9, 0xBA
```

with zeroes appended should be decoded to valid G1 elements. Can either of you @mariano54 or @dfaranha share your thoughts?

Also just for the record, this PR now bumps relic to https://github.com/relic-toolkit/relic/commit/b7b2266a0e4ee6f628f61d3ab638f524a18b52f1. If someone has any objections let me know, i can then drop 1fa53dc91a30b90f007f0991c8c3189dc0e9db43 and 896e33c5210a7272280a72a218d2468a377d97df.

dfaranha:
Currently on holidays, but will catch-up tonight!

mariano54:
I have not reviewed this yet

wjblanke:
ah i saw gene's review and thought it had gone through

hoffmang9:
Sorry - however it did pass Guido's testing and it looked otherwise correct - which is why I approved.

mariano54 (commented):
LGTM

PR #259, opened 2021-07-27
Use CMake's builtin create directory function

AmineKhaldi:
It also gracefully handles the case when a directory already exists.


hoffmang9 (approved):
looks good

PR #276, opened 2021-08-26
Improve and simplify the build process.

AmineKhaldi:


python-bindings/CMakeLists.txt
```diff
@@ -1,8 +1,3 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.14 FATAL_ERROR)
```
hoffmang9, line 1: What is the thought behind dropping the cmake minimum check?

AmineKhaldi: Just that we have it already as the very first line in the top CMakeLists.txt.

hoffmang9 (approved):
AOK

PR #281, opened 2021-09-08
Update relic and tests

mariano54:


AmineKhaldi:
> @AmineKhaldi this can be closed right?

Hmm I believe so, your tests fix commit was already included in PR #277

PR #284, opened 2021-10-02
fix some gcc compiler warnings when using -Wsign-compare

PastaPastaPasta:
util.hpp: In static member function ‘static std::string bls::Util::HexStr(const uint8_t*, size_t)’:
util.hpp:62:25: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
   62 |         for (int i=0; i < len; ++i)


PastaPastaPasta:
ping?

PastaPastaPasta:
ehhem

PR #288, opened 2021-11-03
Provide parameter to FE.zero

niooss-ledger:
In the pure-Python implementation, adding points which lead to the infinity triggers an exception:

    Traceback (most recent call last):
      File "bls-signatures/python-impl/ec.py", line 55, in __add__
        return add_points(self, other, self.ec, self.FE)
      File "bls-signatures/python-impl/ec.py", line 350, in add_points
        return AffinePoint(FE.zero(), FE.zero(), True, ec)
    TypeError: zero() missing 1 required positional argument: 'Q'

This is because `FE.zero` needs to be called with `ec.q`.

Add the missing parameter to the call.


mariano54 (approved):

mariano54:
Thanks!

PR #291, opened 2021-11-13
Updated catch.hpp to 2.13.7 to support glibc 2.34+

llowrey:
Updating catch to 2.13.7 allows it to be built with glibc 2.34 and therefore on Fedora 35. This resolves the issues caused by the breaking change to MINSIGSTKSZ that was made in glibc 2.34.

### DIFF
--- a/contrib/catch/catch.hpp
+++ b/contrib/catch/catch.hpp
@@ -1,9 +1,9 @@
 /*
- *  Catch v2.13.4
- *  Generated: 2020-12-29 14:48:00.116107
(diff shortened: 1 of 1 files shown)


PastaPastaPasta (commented):
utACK, was also unable to build on fedora 35. Applying this patch fixed the build error.

Would like to see this reviewed / merged

hoffmang9:
Closing in favor of https://github.com/Chia-Network/bls-signatures/pull/360

PR #292, opened 2021-11-13
Also build Python 3.10 wheels

altendky:
* Update cibuildwheel to v2.2.2
* Build wheels for 3.10 (except M1)

Draft for:
- [x] consider abi3
  - It isn't apparent that pybind11 supports the stable ABIs.
- [x] break apart
- [x] ~option version in test job names~


hoffmang9 (approved):
aok

hoffmang9 (approved):
aok

PR #293, opened 2021-11-13
Build on main, tags, and PRs

altendky:
As is this builds on any push and also any PR.  This results in double workflows on PRs.

### DIFF
--- a/.github/workflows/build-aarch64.yml
+++ b/.github/workflows/build-aarch64.yml
@@ -1,6 +1,14 @@
 name: Build ARM64 wheel
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - '**'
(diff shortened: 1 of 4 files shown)


altendky:
When reviewing, please consider how releases are done.  I do not know this process and as such don't know how it might affect it.  Mostly this should just stop the double builds and add builds on tags.

cmmarslender (approved):
Seems ok to me - I think the only real change here is just not running on pushes to branches other than main without a PR being open - otherwise, should run in all the existing cases.

altendky:
That, plus it will run on tags.

cmmarslender:
> That, plus it will run on tags.

Should have already been running on tags, since tags is within `push`, and push (with no further restrictions) was already selected before (this is also how releases happen - when the event was for a tag push)

PR #294, opened 2021-11-14
Test all Pythons on Intel Linux

altendky:


altendky:
I suppose the primary test suite is run by cibuildwheel when creating the wheels, so this isn't required for primary version testing coverage.  But, it does seem simple enough to have and is executed in parallel and any extra runner time can be taken out of the runner time saved by https://github.com/Chia-Network/bls-signatures/pull/293.  `:]`  Maybe at some point the pure Python testing should be separated out to avoid duplication.  Or, maybe this isn't worth much.

hoffmang9 (approved):
lgtm

PR #297, opened 2021-12-01
Add bindings for GO

shotonoff:
### Motivation

Since there is no go-bindings for this library, we (**Dash**) decided to add them here.

### Overview

This PR is the implementation of GO bindings for the chia-bls library.

### What was done

There was added bindings for all basic components:
* Schemes (Basic, Aug, Pop)
* PrivateKey
* G1 and G2 elements


hoffmang9:
Happy to get this in once it's passing. Anything we can do to help?

mariano54:
We should probably merge this is before we add any new features

mariano54:
@xdustinface do you think we can merge this?

justinengland:
@xdustinface want to double check you opinion here, can this be merged?

xdustinface:
Seems to be good to be merged yes, if we can keep it maintained and up to date? Even though i worked with this bindings before and im somewhat familiar with them i want to point out thats been a while and im not really into go, it was my only insight into it.

Also i guess, its been a too long time since this was opened, sorry for not replying here earlier.. :/ But we should probably have the go version and dependencies adjusted here, maybe. @shotonoff Any thoughts on this?

xdustinface:
Closing to rerun CI after some recent changes here.

emlowe:
Closing PR - generally since this was moved to use BLST underneath, it's likely easier and better to just use the BLST go bindings directly. Note, chia-blockchain doesn't use this code either as it was migrated into a Rust-based wheel, although that is still using BLST. see https://github.com/supranational/blst

PR #304, opened 2022-01-08
Build Python 3.9 and 3.10 M1 wheels

hoffmang9:


.github/workflows/build-m1-wheel.yml
```diff
@@ -29,29 +29,44 @@ jobs:
...
       run: |
-        arch -arm64 pip install mypy
-        arch -arm64 mypy --config-file mypi.ini python-bindings python-impl
-
-    - name: Build m1 wheels
-      run: |
-        arch -arm64 pip wheel -w dist --no-deps .
-        arch -arm64 delocate-wheel -v dist/*.whl
+        rm -rf venv
+        arch -arm64 /opt/homebrew/opt/python@3.9/libexec/bin/python -m venv venv
```
cmmarslender, line 40:
```suggestion
        arch -arm64 $(brew --prefix python@3.9)/libexec/bin/python -m venv venv
```

I think we've been trying to get the prefix dynamically since it changes to differently locations on Intel vs M1 macs - its easier to port something to work on any of them without necessarily knowing the details of where homebrew puts these things

cmmarslender (changes requested):

.github/workflows/build-m1-wheel.yml
hoffmang9: Smart. Adding it globally.

cmmarslender (approved):

PR #311, opened 2022-01-24
python-bindings: Release the python GIL whenever it makes sense

xdustinface:


python-bindings/pythonbindings.cpp
```diff
@@ -170,32 +175,40 @@ PYBIND11_MODULE(blspy, m)
...
-        .def("sk_to_g1", [](const PrivateKey &seckey){ return AugSchemeMPL().SkToG1(seckey); })
+        .def("sk_to_g1", [](const PrivateKey &seckey){
+            py::gil_scoped_release release;
+            return AugSchemeMPL().SkToG1(seckey);
+        })
         .def(
             "key_gen",
             [](const py::bytes &b) {
+                py::gil_scoped_release release;
                 std::string str(b);
```
mariano54, line 186: Here we are accessing the python object without holding the GIL, is that OK?

python-bindings/pythonbindings.cpp
```diff
@@ -170,32 +175,40 @@ PYBIND11_MODULE(blspy, m)
...
         })
         .def("aggregate", [](const vector<G2Element>& signatures) {
+            py::gil_scoped_release release;
             return AugSchemeMPL().Aggregate(signatures);
         })
         .def(
             "sign",
             [](const PrivateKey &pk, const py::bytes &msg) {
+                py::gil_scoped_release release;
                 std::string s(msg);
```
mariano54, line 212: Same here, and it's similar in other places

mariano54 (commented):
I can't really say whether it's going to crash or not like last time, but we might as well merge it and start testing

python-bindings/pythonbindings.cpp
```diff
@@ -139,8 +143,8 @@ PYBIND11_MODULE(blspy, m)
             [](const G1Element &pk,
                const py::bytes &msg,
                const G2Element &sig) {
+                py::gil_scoped_release release;
                 std::string s(msg);
```
arvidn, line 147: I think you're releasing the GIL too early here. `msg` is a python type, I don't think you can read from it after releasing the GIL.

python-bindings/pythonbindings.cpp
```diff
@@ -99,37 +91,49 @@ PYBIND11_MODULE(blspy, m)
...
-        .def("sk_to_g1", [](const PrivateKey &seckey){ return BasicSchemeMPL().SkToG1(seckey); })
+        .def("sk_to_g1", [](const PrivateKey &seckey){
+            py::gil_scoped_release release;
+            return BasicSchemeMPL().SkToG1(seckey);
+        })
         .def(
             "key_gen",
             [](const py::bytes &b) {
+                py::gil_scoped_release release;
                 std::string str(b);
```
arvidn, line 111: I think copying `b` to `str` needs to happen under the GIL

arvidn (changes requested):
I think you should review all places where we copy from (and mayby into) python objects, and ensure the GIL is held

xdustinface:
@arvidn @mariano54 Good points 👍 See the last force push!

python-bindings/pythonbindings.cpp
```diff
@@ -55,17 +45,19 @@ PYBIND11_MODULE(blspy, m)
                     throw std::invalid_argument(
                         "Length of bytes object not equal to PrivateKey::SIZE");
                 }
+                py::gil_scoped_release release;
                 auto data_ptr = reinterpret_cast<const uint8_t *>(info.ptr);
```
mariano54, line 49: info is still in python right?

python-bindings/pythonbindings.cpp
```diff
@@ -453,6 +488,7 @@ PYBIND11_MODULE(blspy, m)
                 throw std::invalid_argument(
                     "Length of bytes object not equal to G2Element::SIZE");
             }
+            py::gil_scoped_release release;
```
mariano54, line 491: same here with info

python-bindings/pythonbindings.cpp
```diff
@@ -553,6 +602,7 @@ PYBIND11_MODULE(blspy, m)
...
             auto data_ptr = static_cast<uint8_t *>(info.ptr);
             std::vector<uint8_t> data(data_ptr, data_ptr + info.size);
```
mariano54, line 607: and here

python-bindings/pythonbindings.cpp
```diff
@@ -581,6 +632,7 @@ PYBIND11_MODULE(blspy, m)
                     throw std::invalid_argument(
                         "Length of bytes object not equal to GTElement::SIZE");
                 }
+                py::gil_scoped_release release;
```
mariano54, line 635: and here

python-bindings/pythonbindings.cpp
xdustinface: Hmm i assumed that the `PyObject_GetBuffer` in `buffer::request` copies it over but it might not from a quick deeper look at the code.. will review again tomorrow!

xdustinface: Okay changed it, see 353a623. There were even some more cases were `info.ptr` was passed into `Bytes` and then into `::FromBytes` which still would have been an access to that memory without the GIL (`Bytes` only stores the pointer).

mariano54:
Let's get this in after 1.3?

xdustinface:
> Let's get this in after 1.3?

Yeah sure good to me, im anyway waiting for 1.3 that i can start to bug again about all the GIL release PRs in the other repos too :)

mariano54 (approved):

arvidn (approved):
it would be nice to have a stress test where we run a bunch of these operations in a loop in multiple threads simultaneously

PR #329, opened 2022-05-07
lighter-weight buffer handling

arvidn:
make our span-like type (`Buffer`) more span-like and when we need to make copies of (fixed size, small) buffers, make them on the stack rather than in fresh heap allocations.

### DIFF
--- a/src/elements.hpp
+++ b/src/elements.hpp
@@ -40,10 +40,9 @@ public:
         g1_set_infty(p);
     }
 
-    static G1Element FromBytes(const Bytes& bytes);
-    static G1Element FromBytesUnchecked(const Bytes& bytes);
+    static G1Element FromBytes(Bytes bytes);
+    static G1Element FromBytesUnchecked(Bytes bytes);
     static G1Element FromByteVector(const std::vector<uint8_t> &bytevec);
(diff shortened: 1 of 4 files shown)


arvidn:
@mariano54 it's probably best if your PR lands first. I'll rebase this one onto `main` afterwards.

src/elements.hpp
```diff
@@ -82,10 +81,9 @@ class G2Element {
         g2_set_infty(q);
     }
 
-    static G2Element FromBytes(const Bytes& bytes);
-    static G2Element FromBytesUnchecked(const Bytes& bytes);
+    static G2Element FromBytes(Bytes bytes);
+    static G2Element FromBytesUnchecked(Bytes bytes);
     static G2Element FromByteVector(const std::vector<uint8_t> &bytevec);
-    static G2Element FromByteVectorUnchecked(const std::vector<uint8_t> &bytevec);
```
mariano54: Shouldn't we have `FromByteVectorUnchecked` if we have the other 3, for consistency?

mariano54 (commented):
Can you rebase this?

mariano54 (approved):

PR #334, opened 2022-05-31
Fix bit in point deserialization (pure Python)

alessandroguggino:
If not fixed it may cause some issues, e.g. C_bit will be =128 instead of =1 or S_bit =32 instead of =1.
The shift removes the trailing zeros in order to get just a single bit.

### DIFF
--- a/python-impl/ec.py
+++ b/python-impl/ec.py
@@ -238,6 +238,7 @@ def sign_Fq2(element, ec=default_ec_twist) -> bool:
 
 
 def point_to_bytes(point_j: JacobianPoint, ec, FE) -> bytes:
+    # Zcash serialization described in https://datatracker.ietf.org/doc/draft-irtf-cfrg-pairing-friendly-curves/
     point = point_j.to_affine()
     output = bytearray(bytes(point.x))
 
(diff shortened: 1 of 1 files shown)


alessandroguggino:
Hello @mariano54, does it look good to you?

mariano54 (commented):
Have you tested that this is equivalent to the C++ version?

alessandroguggino:
> Have you tested that this is equivalent to the C++ version?

No, I have not tested it in C++ because I am working in pure Python right now.
Since you told me, I looked into it and I saw that there may be the same problem.
The problem arises when we check for the sign bit: `bytes[0] & 0x20` is equal to `0` or `32 (00100000 in bits)` and never to `1`.

Thus, in C++ we have:
https://github.com/Chia-Network/bls-signatures/blob/f35885060491ec3b95afbf000a31e7409019358d/src/elements.cpp#L60
and I think we should modify this line in: `if ((bytes[0] & 0x20) >> 5){` or `if ((bytes[0] & 0x20) == 0x20){`

alessandroguggino:
@mariano54

> Have you tested that this is equivalent to the C++ version?

Yes, I tested the differences between the code in C++ and in pure Python.
I looked into it and I saw that in C++ there is no problem.
The problem arises when we check for the sign bit: `bytes[0] & 0x20` is equal to `0` or `32 (00100000 in bits)` and never to `1`.

In C++ we have:
https://github.com/Chia-Network/bls-signatures/blob/f35885060491ec3b95afbf000a31e7409019358d/src/elements.cpp#L60
and it is not a problem since we can enter in the "if clause" even if we get `bytes[0] & 0x20 = 32`.

While in Python:
https://github.com/Chia-Network/bls-signatures/blob/f35885060491ec3b95afbf000a31e7409019358d/python-impl/ec.py#L300
_(where `S_bit = bytes[0] & 0x20`)_
this is a problem because `sign_fn` returns `True` or `False`, so we can never enter in the "if clause" when `True == 32`. 

Thus, my solution is to have `S_bit` equal to `0` or `1`, removing the trailing zeros with the shift (`>>`).

mariano54 (approved):

PR #338, opened 2022-07-10
rework build matrix to a single workflow

altendky:
This closely follows the form of https://github.com/Chia-Network/chiavdf/pull/118.

Ideas:
- [ ] Can we centralize the build of gmp?
- [ ] Can we agree on either `src/lib/` or `contrib/` between here and chiavdf?

Draft for:
- [x] We are not patching with `src/lib/gmp-patch-6.2.1/mpz/inp_raw.c`, is that ok?
- [x] Are the sha256 hashes required?  I don't recall seeing them elsewhere.


hoffmang9:
The hashes were a start at the upstream repos confirming them. The main repo is publishing a package hash now and we're pinning to version so probably ok to drop for now

altendky:
I have no problem putting them in, but in so much as we 'need' them we should do it everywhere I would think.  I'll leave them out for now but would be happy to bring them back someday.  Maybe as an action to process a GitHub Actions artifact of all the packages?

cmmarslender (approved):

PR #341, opened 2022-08-04
Gt element unchecked

arvidn:
We sometimes serialize and deserialize `GTElements` to pass them across process boundaries. Specifically, when the mempool validates a transaction, it does so in a separate process and receives the new BLS cache entries back from that process (in serialized form). Since we trust the other process (it's also part of the full node), we don't have to validate the `GTElement` when inserting it into the main BLS cache.

This overload adds symmetry with `G1Element.from_bytes_unchecked()` and `G2Element.from_bytes_unchecked()`


mariano54 (approved):

arvidn:
tests for `GTElement` are lacking. that should be addressed, perhaps in this PR.

PR #343, opened 2022-09-03
Add flake for nix build support

abueide:
This PR depends on the flake from the relic fork being merged in: https://github.com/Chia-Network/relic/pull/3

Since nix doesn't support downloading dependencies during the build process, Findrelic.cmake was added, and a flag BUILD_LOCAL was added. If BUILD_LOCAL is enabled, it will search for dependencies only on the local machine rather than using FetchContent to support nix's deterministic build system.

If BUILD_LOCAL is not specified or is false, build behavior is unchanged.


justinengland:
hello! Can you please resolve the conflict?

PR #344, opened 2022-09-16
Allow AffinePoint to be correctly turned into bytes

ayrtondutchman:
Yesterday I was messing around with bls-signatures python impl and bindings and noticed that in the python impl when a AffinePoint is turned into bytes it doesn't handle it like it handles a JacobianPoint turned into bytes. Even though the process is exactly the same as the first thing that is done to the Jacobian is it is turned into Affine.

### DIFF
--- a/python-impl/ec.py
+++ b/python-impl/ec.py
@@ -84,6 +84,9 @@ class AffinePoint:
             + ")\n"
         )
 
+    def __bytes__(self) -> bytes:
+        return point_to_bytes(self, self.ec, self.FE)
+
     def __eq__(self, other) -> bool:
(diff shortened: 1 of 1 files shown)


mariano54 (approved):

ayrtondutchman:
@mariano54 thanks for reviewing. I saw that it couldn't be merged because of something called vigilant mode. Which was new to me. Now that I have read up about it. I think I have been able to sign the older commit. (Was a bit of a hassle as you can see) Now I was wondering if the workflows could get re-approved.

hoffmang9 (approved):
lgtm now that it’s signed

PR #358, opened 2023-02-23
fix: fixed UB with reading unitialized data

knst:
This call `memcmp` is quite useless, because result is not used and move over reading un-itialized data.


arvidn (approved):

src/test.cpp
```diff
@@ -56,7 +56,6 @@ void TestHKDF(string ikm_hex, string salt_hex, string info_hex, string prk_expec
 
 TEST_CASE("class PrivateKey") {
     uint8_t buffer[PrivateKey::PRIVATE_KEY_SIZE];
-    memcmp(buffer, getRandomSeed().data(), PrivateKey::PRIVATE_KEY_SIZE);
```
arvidn, line 59: I believe this was intended to say `memcpy()`

arvidn (changes requested):

PR #359, opened 2023-02-23
fix: add missing destructor for CoreMPL and `final` for nested classes

knst:
There is no actual memory leak because no data in the the CoreMPL

But it resolves clang's warning:

    /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/unique_ptr.h:95:2: warning: delete called on non-final 'bls::CoreMPL' that has virtual functions but non-virtual destructor [-Wdelete-non-abstract-non-virtual-dtor]
            delete __ptr;
            ^
    /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/unique_ptr.h:396:4: note: in instantiation of member function 'std::default_delete<bls::CoreMPL>::operator()' requested here
              get_deleter()(std::move(__ptr));
              ^
    main.cpp: note: in instantiation of member function 'std::unique_ptr<bls::CoreMPL>::~unique_ptr' requested here
    static const std::unique_ptr<bls::CoreMPL> pSchemeMLP{std::make_unique<bls::BasicSchemeMPL>()};

### DIFF
--- a/src/schemes.hpp
+++ b/src/schemes.hpp
@@ -39,6 +39,7 @@ class CoreMPL {
 public:
     CoreMPL() = delete;
     CoreMPL(const std::string& strId) : strCiphersuiteId(strId) {}
+    virtual ~CoreMPL() {}
     // Generates a private key from a seed, similar to HD key generation
     // (hashes the seed), and reduces it mod the group order
     virtual PrivateKey KeyGen(const vector<uint8_t>& seed);
@@ -96,7 +97,7 @@ protected:
     bool NativeVerify(g1_t *pubKeys, g2_t *mappedHashes, size_t length);
 };
 
-class BasicSchemeMPL : public CoreMPL {
+class BasicSchemeMPL final : public CoreMPL {
 public:
     static const std::string CIPHERSUITE_ID;
     BasicSchemeMPL() : CoreMPL(BasicSchemeMPL::CIPHERSUITE_ID) {}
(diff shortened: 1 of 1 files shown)


arvidn (approved):

src/schemes.hpp
```diff
@@ -39,6 +39,7 @@ class CoreMPL {
 public:
     CoreMPL() = delete;
     CoreMPL(const std::string& strId) : strCiphersuiteId(strId) {}
+    virtual ~CoreMPL() {}
```
arvidn, line 42: is this still necessary when you make the other types final?

knst: Yes, because final in other class mean just no more inheritance and it doesn't need it's own virtual destructor. But above class/struct still needs so

arvidn: ok. I guess we already have other virtual functions in this class, and we *do* use it polymorphically. so this seems right.

wallentx:
Hi @knst, our branch protection rules require that all commits be signed before merging. If you know how to retroactively sign your initial commit, we can proceed with this PR. If you aren't sure how to do that, I've also prepared this PR https://github.com/Chia-Network/bls-signatures/pull/371 to merge in, with your attribution listed as co-author.

Please let me know how you'd like me to continue 👍

wallentx:
@knst I'm going to go ahead and close out this PR, and merge in your changes in #371

PR #364, opened 2023-04-21
feat: implement memory-copy only of G1Element and G2Element

PastaPastaPasta:
This implements a performant memory-only copy for G1Element and G2Element; I don't have benchmarks, but when we used this we saw significant performance improvements.

### DIFF
--- a/src/elements.hpp
+++ b/src/elements.hpp
@@ -59,6 +59,7 @@ public:
     GTElement Pair(const G2Element &b) const;
     uint32_t GetFingerprint() const;
     std::vector<uint8_t> Serialize() const;
+    G1Element Copy();
 
     friend bool operator==(const G1Element &a, const G1Element &b);
     friend bool operator!=(const G1Element &a, const G1Element &b);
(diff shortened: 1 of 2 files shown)


hoffmang9:
Now that we've completed some needed housekeeping, could you rebase so we can get a clean CI and get this in?

Thanks very much for your upstream!

-Gene

src/elements.hpp
```diff
@@ -59,6 +59,7 @@ class G1Element {
     GTElement Pair(const G2Element &b) const;
     uint32_t GetFingerprint() const;
     std::vector<uint8_t> Serialize() const;
+    G1Element Copy();
```
arvidn, line 62: in C++, this functionality is implemented by copy constructors. I can't see a good reason to have this be a member function instead. But, if there is a good reason, it should *at least* be a `const` member function

arvidn (changes requested):

emlowe:
Closing PR - this isn't required due to the implicit C++ copy constructors, which are already in memory (essentially a memcpy)

PR #367, opened 2023-05-02
Streamline the processes of building libbls, testing it, and creating the blspy wheels

AmineKhaldi:
* Speedup the process of creating wheels by skipping the compiling and linking of `libbls`' tests and benchmarks.
* We no longer need Windows-specific `CMake` version checking in `setup.py`.
* We no longer need Windows-specific architecture flag passing in `setup.py`.
* We no longer treat Windows in a special way w.r.t. compiling and linking in `setup.py`.
* We no longer need Windows-specific preparations before starting the process of creating wheels.
* We no longer need Windows-specific repairs after wheels creation.
* We no longer need the `relic_ietf_64` repository and its related tasks.

Dedicated to @wjblanke

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ if(DEFINED ENV{RELIC_MAIN})
   set(RELIC_REPOSITORY "https://github.com/relic-toolkit/relic.git")
 else()
   # This is currently anchored to upstream aecdcae7956f542fbee2392c1f0feb0a8ac41dc5
-  set(RELIC_GIT_TAG "215c69966cb78b255995f0ee9c86bbbb41c3c42b")
+  set(RELIC_GIT_TAG "2a4ed8ded3a9a5ac5e39f3877f357264b2e16f11")
   set(RELIC_REPOSITORY "https://github.com/Chia-Network/relic.git")
 endif()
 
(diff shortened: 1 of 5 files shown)


hoffmang9 (approved):
lgtm

CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ if(DEFINED ENV{RELIC_MAIN})
   set(RELIC_REPOSITORY "https://github.com/relic-toolkit/relic.git")
 else()
   # This is currently anchored to upstream aecdcae7956f542fbee2392c1f0feb0a8ac41dc5
-  set(RELIC_GIT_TAG "215c69966cb78b255995f0ee9c86bbbb41c3c42b")
+  set(RELIC_GIT_TAG "2a4ed8ded3a9a5ac5e39f3877f357264b2e16f11")
```
arvidn, line 50: what changed here? We don't have any regression tests for the BLS functionality itself (afaik), so we should be careful when updating relic to ensure it's still running regression tests

CMakeLists.txt
```diff
@@ -141,6 +154,17 @@ set(PP_METHD "LAZYR;OATEP" CACHE STRING "")
 
 FetchContent_MakeAvailable(relic)
 
+function(handle_mpir_dlls _target)
+  file(GLOB chia_mpir_dlls ${chia_mpir_SOURCE_DIR}/*.dll)
```
arvidn, line 158: `GLOB` makes me nervous. Do we not know what the filename is of this DLL?
The reason it makes me nervous is that it makes the build less deterministic, it depends on what files happen to be on your filesystem. perhaps you have some old files left-over from a previous build or a previous version

CMakeLists.txt
```diff
@@ -63,12 +63,25 @@ FetchContent_Declare(
...
-find_package(gmp)
-if (GMP_FOUND)
-  message(STATUS "Found libgmp")
+if(MSVC)
   set(ARITH "gmp" CACHE STRING "")
+  # We'll be using our custom MPIR build instead of GMP for MSVC builds
+  FetchContent_Declare(
+    chia_mpir
+    GIT_REPOSITORY https://github.com/Chia-Network/mpir_gc_x64.git
+    GIT_SHALLOW true
```
arvidn, line 72: do we want to pin this to a specific commit?
I would think it would make for fewer moving parts as we can update our MPIR build independently of taking that update in the blspy build

arvidn (commented):
looks good generally. just a few comments

CMakeLists.txt
AmineKhaldi: You can check the changes [here](https://github.com/Chia-Network/relic/commit/2a4ed8ded3a9a5ac5e39f3877f357264b2e16f11), and we run tests (just like upstream) in addition to our own bls-signatures tests so we should be reasonably careful (at least to the degree that our tests cover).

CMakeLists.txt
AmineKhaldi: This is a build dependency so in order to alter the state of that GLOB you would need to manually intervene inside the CMake build's _deps folder and add/remove dlls from our fetched MPIR dependency there. I don't mind making this list manual if that makes us feel better, with that said.

CMakeLists.txt
AmineKhaldi: We're in control of this repo, we rarely update those DLLs and we would like to take the latest. Furthermore, having inconsistency in what dlls we fetch from this dependency, across repos, would result in issues.

arvidn (approved):

PR #370, opened 2023-05-08
Fix a typo in class PrivateKey testcase

AmineKhaldi:
Thanks to @knst for bringing attention to it in #358

### DIFF
--- a/src/test.cpp
+++ b/src/test.cpp
@@ -56,7 +56,7 @@ void TestHKDF(string ikm_hex, string salt_hex, string info_hex, string prk_expec
 
 TEST_CASE("class PrivateKey") {
     uint8_t buffer[PrivateKey::PRIVATE_KEY_SIZE];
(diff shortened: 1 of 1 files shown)


src/test.cpp
```diff
@@ -56,7 +56,7 @@ void TestHKDF(string ikm_hex, string salt_hex, string info_hex, string prk_expec
 
 TEST_CASE("class PrivateKey") {
     uint8_t buffer[PrivateKey::PRIVATE_KEY_SIZE];
-    memcmp(buffer, getRandomSeed().data(), PrivateKey::PRIVATE_KEY_SIZE);
+    memcpy(buffer, getRandomSeed().data(), PrivateKey::PRIVATE_KEY_SIZE);
```
knst, line 59: It copies biffer to getRandomeSeed, doesn't it?

arvidn: no. https://man7.org/linux/man-pages/man3/memcpy.3.html

arvidn (approved):

PR #371, opened 2023-05-08
fix: add missing destructor for CoreMPL and final for nested classes

wallentx:
Signing commit from @knst from https://github.com/Chia-Network/bls-signatures/pull/359


hoffmang9 (approved):
@arvidn also approved #359

PR #385, opened 2023-06-08
Fix FromMessage to use hash and not encode

emlowe:
we always hash to point


wjblanke (approved):
ok but need to add the test back in

emlowe:
Closing as this was included in PR 386.

PR #390, opened 2023-06-12
Add GTElement serialization with memcpy

emlowe:
Implement GTElement serialization with straightforward memcpy

Add a few tests to the C++ test code for GTElement


wjblanke (approved):
aok

PR #391, opened 2023-06-12
Use _add_or_double instead of just _add and pin BLST to specific revision

emlowe:
Make sure to use`_add_or_double` rather than just `_add`. This properly supports creating an aggregated signature with the same sig.

Added test case that tests this functionality

Also, pin BLST to revision `a8cd361c9f671577aeab3f074098443af92a53fc` as very recent changes in `origin/master` cause some compilation issues.


emlowe:
close and reopen for CI strangeness

wjblanke (approved):
aok

PR #395, opened 2023-06-14
Use sizeof(blst_fp12) for size of GTElement serialization

emlowe:
Use sizeof(blst_fp12) instead of hard coded number for GTElement::Size


wjblanke (approved):
aok

arvidn (approved):

PR #412, opened 2023-06-21
Add some assembly source files that we'd need for the upcoming nightly build

AmineKhaldi:
This assists #410.


arvidn:
is this supposed to fix this link error? it's not obvious that it would:
https://github.com/Chia-Network/bls-signatures/actions/runs/5329021747/jobs/9654285191?pr=410#step:3:233

AmineKhaldi:
> is this supposed to fix this link error? it's not obvious that it would: https://github.com/Chia-Network/bls-signatures/actions/runs/5329021747/jobs/9654285191?pr=410#step:3:233

It shouldn't, this is related to the Windows side (where these asm files are used).

hoffmang9 (approved):
lgtm

PR #416, opened 2023-06-22
Update CMakeLists.txt to current blst/master

hoffmang9:
Since we now have things building with blst/origin/master, lets pin to current.


wjblanke (approved):
aok

PR #425, opened 2023-08-01
Update to cibuildwheel 2.11.2

emlowe:
Consolidating wheels to cibuildwheel 2.11.2

This is the last version that uses bash in windows, which is needed for some scripts. The scripts could be edited to work in powershell, but this is left for another day


emlowe:
close and reopen for CI

PR #426, opened 2023-08-04
Use pyproject.toml instead of envvars for cibuildwheel

emlowe:
Puts most `cibuildwheel` configuration into `pyproject.toml` instead of using envvars in the github actions.

The rationale is that this makes it a little bit easier to build the wheel locally for debugging and testing as there is less configuration specific to GH actions

Note some of the envvars refer to GH action matrix items, so those remain in the GH workflow.


emlowe:
close and re-open for CI

hoffmang9 (approved):
lgtm

wallentx (approved):

PR #429, opened 2023-09-01
Build and test wheel for riscv64 using QEMU

emlowe:
Set to compile BLST with __BLST_NO_ASM__ on riscv64.

This could be changed to add this setting on anything not x64 or arm - but doing it this way for now


hoffmang9 (approved):
What? No riscv64 custom runner to test with? 🤣

emlowe:
QEMU riscv emulation is possible

hoffmang9:
You might guilt me into building you that flow..

cmmarslender (approved):

PR #438, opened 2023-10-25
Division operator for GTElement (e1/e2)

lokingdav:
Currently creating a [witness encryption scheme](https://github.com/kofi-dalvik/bls-witness-encryption) based on BLS signatures. I am using this library. However, I needed the GTElement class to support inversing through divisions. Instead of reimplementing element classes (since r is private in GT), this PR fixed the issue. 

I know BLS signature implementation does not use division, but libraries that depend on this might need it. Moreover, my witness encryption library currently depends on my fork, and I want to use the main Chia-Network/bls-signatures instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,12 @@ Implements BLS signatures with aggregation using [blst library](https://github.c
 for cryptographic primitives (pairings, EC, hashing) according to the
 [IETF BLS RFC](https://datatracker.ietf.org/doc/draft-irtf-cfrg-bls-signature/)
 with [these curve parameters](https://datatracker.ietf.org/doc/draft-irtf-cfrg-pairing-friendly-curves/)
-for BLS12-381.
+for BLS12-381. The blst library has been [audited](https://medium.com/supranational/introducing-blst-2b6a988d68ee).
 
 Features:
 
 * Non-interactive signature aggregation following IETF specification
-* Works on Windows, Mac, Linux, BSD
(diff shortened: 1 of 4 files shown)


hoffmang9 (approved):
Thanks for your contribution

hoffmang9:
Can you sign your commit?

kofi-dalvik:
Thanks @hoffmang9 , 
Signed now

hoffmang9:
image
Looks like something is still a bit off.

kofi-dalvik:
> image Looks like something is still a bit off.

Yeah i noticed and I'm resolving it. Used a different email for this repository

emlowe:
Apologies for the long delay in looking at this. But I have a question. The included test is odd since division is not normally commutative - so one doesn't expect a test where `a/b == b/a`

I would expect something like `(a*b)/b == a` - does this version of division pass that test?

kofi-dalvik:
Hi @emlowe thanks for this message. In fact, I forgot about this PR. 

Your observation is correct, which would be a more precise test. The current test works because gt1 and gt2 are both GT generators and are the same. It will make more sense that gt1 and gt2 are different elements. I'll update the test.

emlowe (changes requested):
I'm not sure it's correct to call "multiplication of the inverse" a division operation

emlowe:
closing PR

PR #447, opened 2024-01-22
Add python 3.12

emlowe:
Add python 3.12 to matrix
remove old exclude of python 3.7 on macOS
upgrade cibuildwheel to 2.16.2 (current latest) needed for python 3.12 support


hoffmang9 (approved):
lgtm

PR #448, opened 2024-01-24
Update blsjs.d.ts

BubbleTrouble14:
Updated blsjs.d.ts type inside the G1Element generator to return G1Element and not G2Element, or is this already correct ?


ChiaMineJP (changes requested):
Thank you for your PR.
`blsjs.d.ts` is just a generated code so you need to update the original cpp file which generates this declaration file.

https://github.com/Chia-Network/bls-signatures/blob/main/js-bindings/jsbindings.cpp#L67

BubbleTrouble14:
Ah sorry. I'll close this pull request and open a new one with the correct fix.

PR #455, opened 2024-03-24
Update jsbindings.cpp

BubbleTrouble14:
Changed in the correct place now.


ChiaMineJP (commented):
I'd suggest to build it as described in
https://github.com/Chia-Network/bls-signatures/blob/main/js-bindings/README.md#build

emlowe:
This can't be correct - you change the G2Element to use the G1Element generator?

BubbleTrouble14:
> This can't be correct - you change the G2Element to use the G1Element generator?

Yes indeed its wrong was supposed to update the G1Element to use the G1Element generator. But hadn‘t got round to then building it and doing it correctly. As when looking at the current implementation the g1 generator returns the G2Element in the types which i assume is incorrect ?

BubbleTrouble14:
[Location](https://github.com/Chia-Network/bls-signatures/blob/main/js-bindings/blsjs.d.ts#L44) in the generated types as seen here.

PR #469, opened 2024-10-22
Bump pybind for cmake deprecation

hoffmang9:


hoffmang9:
pybind11 is not the issue - looks like something relating to Catch2

PR #471, opened 2024-10-22
install python for blst nightly checks before building instead of after

altendky:
https://github.com/Chia-Network/bls-signatures/actions/runs/11468270388


hoffmang9:
lgtm but seems I can't approve anymore

hoffmang9 (approved):
Must have been an issue on mobile. Approved

